PR #444, opened 2014-11-28
Fix less variable overrides by executing less-modify.js between the build and saving

### LINKED ISSUE #443: Less variable overrides are not always applied during CSS build
@vangheem I'm puzzled. When I set a LESS variable through resources registry control panel, in the development mode it's properly applied for all bundles. When I build the bundles (TTW) for production mode, variables are applied at least for the plone-bundle, but not my barceloneta-derived bundle.

I looked into LESS->CSS-compilation in resourceregistry-pattern and it looked quite simple: all bundle builds get the same less-variables.js before less-compilation.

Any idea, what could be so much different in barcelona.plone.less when compared to plone.less so that variable-overrides are applied in development mode, but not in build?

datakurre:
See also #443
[fixes #443]

### DIFF
--- a/mockup/patterns/resourceregistry/js/builder.js
+++ b/mockup/patterns/resourceregistry/js/builder.js
@@ -113,6 +113,10 @@ define([
       /* XXX okay, wish there were a better way,
          but we need to pool to find the out if it's down loading less */
       self.addResult(config.less.length + _t(' css files to build'));
+      var lessModified = Boolean(
+          self.rview.options.data.lessModifyUrl === null ||
+          self.rview.options.data.lessModifyUrl === undefined
+      );
       var checkFinished = function(){
         var $styles =  $('style[type="text/css"][id]', iframe.document);
         for(var i=0; i<$styles.length; i=i+1){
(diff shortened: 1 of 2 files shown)


jensens:
seems this needs a rebase with master. also @bloodbare can you look at this?

datakurre:
I rebased, but don't have a current Plone 5 project to test this with. Yet, the issue should be easily confirmed as in #443.

mockup/patterns/resourceregistry/js/builder.js
```diff
@@ -150,6 +154,16 @@ define([
               self.finished(true);
             }
           });
+        }else if($styles.length === config.less.length){
+          $styles.each(function(){$(this).remove();});
+
+          script = document.createElement('script');
+          script.setAttribute('type', 'text/javascript');
+          script.setAttribute('src', self.rview.options.data.lessModifyUrl);
```
vangheem: I looked at this code and it just seemed funny. At this point, wouldn't self.rview.options.data.lessModifyUrl be null/undefined? So why are we loading it this way?

vangheem:
I read the two tickets and I understand what is going on now. Thank you for catching this.

PR #457, opened 2015-01-13
Resource registry: show bundle compilation metadata

thet:
I was wondering, how someone can find out if and when a bundle was compiled.

Therefore: In the resource registry bundle detail view, add the fields last_compilation, jscompilation and csscompilation for display.  This gives more insight about the state of each bundle.

plus a small fix in the ``_buildCSSBundle`` iframe template.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -26,6 +26,11 @@ New patterns:
 
 Fixes and enhancements:
 
+- In the resource registry bundle detail view, add the fields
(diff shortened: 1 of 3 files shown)


thet:
i think, it's much less confusing to let these fields non-editable. otherwise people would think, they have to edit it. i think these options cover all use cases:
- compiling through-the-web, where the compilation date and compiled bundle paths are auto-generated
- compiling via a grunt task and importing via generic setup, where you can set the compilation date and bundle paths to whatever you like.

PR #458, opened 2015-01-21
jsi18n -> plonejsi18n

kiorky:
This refs plone/plone.app.widgets@4b397c3

### DIFF
--- a/mockup/js/i18n-wrapper.js
+++ b/mockup/js/i18n-wrapper.js
@@ -2,9 +2,9 @@
  *
  * This is a singleton.
  * Configuration is done on the body tag data-i18ncatalogurl attribute
- *     <body data-i18ncatalogurl="/jsi18n">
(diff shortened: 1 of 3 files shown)


kiorky:
cc @ebrehault @datakurre  @vangheem

datakurre:
+1

This shouldn't hurt anyone, because AFAIK in Plone 5 p.a.widgets should be merged into CMFPlone anyway.

vangheem:
This is fine. I'm not sure I understand why the change is necessary though.

vangheem:
and that test failure doesn't make any sense.

datakurre:
@vangheem For Plone 4.3, which does not jsi18n info in its p.a.layout version.

vangheem:
I see. That is fine. FWIW, I'm not worrying about 4.3 for now on with mockup and plone.app.widgets. It's been holding us back and too much work to try and make everything work both ways. If someone else wants to try, that's fine.

PR #459, opened 2015-01-22
Makefile rework

benniboy:


vangheem:
hmmm, I can merge but the upgrade causes a few more test failures. Are you getting those test failures when running locally?

benniboy:
Locally the tests run fine. I reverted commit f4d9269 because otherwise i couldn't run "make". I don't know why on travis the build fails :/

vangheem:
Ahh, I didn't even know that was reverted--how nice.

I suppose we might have to attempt testing specifically with the failed browsers and see what's going on.

benniboy:
I think commit f4d9269 was made to get the tests running on travis, but it's odd that the same settings don't run locally. I haven't had such a problem before.

vangheem:
Nah, that was in response to the plone jenkins CI having a test failure. If a test fails, often things just start getting reverted. This test had nothing to do with the jenkins test failure.

benniboy:
ah okay =)

vangheem:
I'm worried about these that seem like test regressions here: https://travis-ci.org/plone/mockup/jobs/47911616#L1696

jcbrand:
@vangheem While working on tests in the patternslib-merge branch I figured out why those tests fail.

It has to do with the fact that the "focus" event is no longer being catched, at least for me in Chromium 35.0.1916.153.

These commits fixed them for me:
https://github.com/plone/mockup/commit/236c15f115ff412589e55e7a79666876fd8ea90a
https://github.com/plone/mockup/commit/24235ba1b0477a355f61857724745e863fe6d024

The 2nd commit only makes sense when we use the Patternslib registry.

vangheem:
thanks @jcbrand

PR #460, opened 2015-01-24
Patternslib merge: Use Patternslib's scanner and registry

jcbrand:
**Don't merge yet, first we need to merge the pull request in mockup-core.**
https://github.com/plone/mockup-core/pull/25

During the Alpine sprint I modified Patternslib, mockup and mockup-core to enable Mockup to use the Patternslib scanner and registry.

This allows us to:

* Use Patternslib patterns with Mockup/Plone.
* Use Mockup patterns with Patternslib outside of Plone.

### Changes required to individual patterns from now on.

The integration with Patternslib require that some small changes be made to newly developed Mockup patterns:

1. Patterns should now use ``pat-registry`` as dependency instead of ``mockup-registry``.

    ``` javascript
    define([
      'jquery'
      'mockup-patterns-base',
      'pat-registry'
    ], function($, Base, registry) {
    ```
2. Patterns' selectors are now explicitly specified via the *trigger* attribute.

    For example:
    ``` javascript
      var Modal = Base.extend({
        name: 'modal',
        trigger: '.pat-modal',
    ````
3. Because of change 2, patterns now fire events via the *emit* method, instead of the *trigger* method.


vangheem:
Awesome. Unless you want more review, please go ahead and merge when you're ready.

.travis.yml
```diff
@@ -41,4 +41,4 @@ matrix:
     - env: BROWSERS="SL_Android"
 branches:
   only:
-    - master
+    - patternslib-merge
```
vangheem: This should be switched back before merge.

mockup/bower.json
```diff
@@ -14,15 +14,21 @@
     "jquery-form": "3.46.0",
     "jquery.cookie": "1.4.1",
     "lodash": "2.4.1",
-    "mockup-core": "1.2.16",
+    "logging": "",
+    "mockup-core": "#patternslib-merge",
```
vangheem: Again, switch this to released mockup-core merge when done.

mockup/patterns/tinymce/js/links.js
```diff
@@ -582,41 +600,42 @@ define([
...
       } else {
-        self.linkType = 'externalImage';
-        self.linkTypes.externalImage.set(src);
+        this.linkType = 'externalImage';
+        this.linkTypes.externalImage.set(src);
       }
     },
+
     guessAnchorLink: function(href) {
-      var self = this;
```
vangheem: FWIW, I like the style of setting `var self = this;`. I know it's not necessary all the time but other times, it prevents confusing when using with inline functions.

jcbrand: To be honest, I don't particularly like it. It looks to me like a Python developer trying to write Javascript.

Generally, the consensus in the Javascript community appears to be to use ``that`` as variable name if you want to refer to the context enclosing a callback (i.e instead of ``self``). This is what Douglas Crockford suggests in "Javascript, the good parts".

In general however, I prefer using things like ``function.prototype.bind`` or ``$.proxy`` to explicitly set the context of ``this`` for callback functions.

mockup/bower.json
jcbrand: Thanks, will do.

PR #461, opened 2015-02-06
TinyMCE lang

ebrehault:
- Use `i18n.currentLanguage` to initialise TinyMCE lang option,
- Fallback to closest lang if the required one is missing in TinyMCE (for instance, if `fr_be.js` is missing, we try `fr.js` and if `fr.js` is missing, we try `fr_Fr.js`)


ebrehault:
@bloodbare, have a look when you have a moment please.

mockup/patterns/tinymce/pattern.js
```diff
@@ -310,35 +350,34 @@ define([
...
+          self.options.scales = _.map(self.options.scales.split(','), function(scale){
+            var scale = scale.split(':');
+            return {
+              part: scale[1],
+              name: scale[1],
+              label: scale[0]
+            }
+          });
+        }
+        if(typeof(self.options.folderTypes) === 'string'){
```
thet: shouldn't that and the following code be outside the current block and not within ``initLanguage``?

vangheem:
Maybe bad merge? This might need to be updated with master.

ebrehault:
@thet @vangheem we need to be done into a callback because `initLanguage` makes an ajax call to the lang file.

thet:
@ebrehault ah, i see. it's a callback function. trying out locally....

PR #462, opened 2015-02-11
Bugfix. Prevent patterns from being loaded multiple times.

jcbrand:
For reference, see my comment here:
https://github.com/Patternslib/Patterns/issues/396#issuecomment-73722048

If we have multiple bundles that all pull in pat-registry and then call
registry.scan, then patterns will be scanned and initialized multiple times.

Instead, we need to call registry.init *if* it's not been called before. If
it's been called before, we don't need to do anything because the patterns will
automatically be registered as soon as they are loaded by require.js and evaluated
by the JS engine and then automatically scanned by the registry.

Note: this change relies on the Patternslib registry and won't work with the fork @vangheem made of pat-registry. Hopefully that fork can now be discarded.

### DIFF
--- a/mockup/bower.json
+++ b/mockup/bower.json
@@ -15,7 +15,7 @@
     "jquery.cookie": "1.4.1",
     "lodash": "2.4.1",
     "logging": "",
-    "mockup-core": "2.1.1",
+    "mockup-core": "git@github.com:plone/mockup-core.git#registry-fix",
     "moment": "2.8.3",
     "pickadate": "3.4.0",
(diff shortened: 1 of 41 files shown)


vangheem:
All files would need to be changed from mockup-registry back to pat-registry.

jcbrand:
> All files would need to be changed from mockup-registry back to pat-registry.

done.

jcbrand:
Travis fails because it cannot check out the "registry-fix" branch or mockup core, which this change depends on.

Tests pass locally.

PR #463, opened 2015-02-12
Thememapper bundle

ebrehault:
Define the Thememapper bundle, used in plone.app.theming.


vangheem:
This was manually merged.

Issue #442, opened 2014-11-27
introduce icons to relateditems vocabulary list


petschki:
I've came across this in my recent project and think, this would be a nice little feature for relateditems list. I've added "getIcon" to the RelatedItems.attributes and updated resultTemplate and selectionTemplate in the pattern to show the icon.

So if one is cool and adds a custom "getIcon" indexer you can enable image preview in relateditems selector like this:
![bildschirmfoto 2014-11-27 um 10 55 26](https://cloud.githubusercontent.com/assets/511761/5215145/76b6c276-7624-11e4-9e84-34c45de3ae28.png)

just thinking loud :)

PR #465, opened 2015-02-20
add content icon to relateditems pattern

petschki:
see #442

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -22,6 +22,9 @@ Changelog
   mockup/GETTING_STARTED.md . Refs: #460.
   [jcbrand]
 
+- Add icons to relateditems pattern (see https://github.com/plone/mockup/issues/442)
+  [petschki]
+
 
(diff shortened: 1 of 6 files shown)


vangheem:
How does this work OOTB? Does plone give a meaningful value for getIcon all the time. Should those img tags be conditionally in the template?

davisagli:
It only returns a value for types that have an icon_expr set in portal_types. Most of the core types don't have an icon there (they get icons via CSS) but some addons still do. In plone.app.contenttypes file and image have a special getIcon indexer which gets an icon from the mimetypes_registry based on the file mimetype

petschki:
So this is what the latest change brings OOTB:
bildschirmfoto 2015-02-23 um 10 30 13

vangheem:
I'm okay with this. This way you can have customized icons thumbnails and icons are the fallback.

Side note: We should match the icons used in Plone 5(defined as only css rules) back into the filesystem icon image files defined.

PR #467, opened 2015-03-08
split into mockup and plone.mockup

thet:
Split out the Python package into ``plone.mockup`` and make ``mockup`` installable via ``bower`` again.


jcbrand:
:thumbsup:

jcbrand:
> If mockup is not anymore a egg its not checked out and developed, so we can't track it on jenkins @tisto as it should not be on checkouts.cfg

How does this work with mockup-core, which AFAIK is not an egg either?

I'm definitely +1 for making mockup a bower package. The advantage for this is to allow other projects to use the mockup patterns.

For example, Plone Intranet doesn't (yet?) use the Plone 5 resource registries and instead manages JS dependencies manually via NPM and Bower.

bloodbare:
Every body wants that mockup is a bower component, and its ok! Just we need to figure out how to do it. Mockup core is a really small peace of code that its embedded on CMFPlone as the other resources using bower. Using mockup-core as an external library in my opinion is ok as its so small that its stability is easy to maintain. Mockup as itself are the widgets on top of this library and they belong to the actual development process, as mockup-core should remain stable and not changing a lot. is it ok ?

jcbrand:
Interesting that you bring up "development process".  This is perhaps something that needs to be clarified.

Patternslib was inspired by a  "design first" methodology whereby a static, patterns-enhanced prototype (or Diazo theme if you like) is initially designed, approved by the customer (or community) and then integrated into the website. This is an iterative, rapid-prototyping process whereby the client gets to see what the site will look like even before any Plone development has been done.

Patternslib is primarily meant to make designer's lives easier, not developers. It also helps people selling Plone sites to show clients how things will look and behave without having to first write the backend functionality.

The is the methodology we are currently following with Plone Intranet so it's something that more and more people are becoming acquainted with.

I'm not saying this is the holy grail of development methodologies, just that this is the main point of patterns, and that patterns lose a lot of their power when not used with such a "design first" approach.

This is apparently not the approach that was taken with mockup and I get the impression that many mockup devs are not aware of this "design first" philosophy that is the inspiration behind patterns.

One consequence of the "design first" approach is that it's possible to do the majority of pattern development completely outside of Plone. For example, we're currently having the Plone Intranet sprint where the bulk of the patterns work is done in the standalone patterns with just the Node development stack. No Plone involved.

The designer(s) then uses these patterns to create the prototype which then gets applied to the site via Diazo.

IMHO, this should be the same with mockup. It should be possible to work on mockup patterns completely independently of Plone. Yes, it will still be required to test whether the patterns work in Plone, or with the Plone backend, but that's not where the main development would then happen.

So, IMHO, if you need to have a full Plone site to do pure mockup pattern development, then you're doing it wrong.

vangheem:
This is incorrect. You can do mockup dev without having a full plone site. Of course, that was and still is a goal of mockup. All @bloodbare is saying is that mockup is also a plone project and tightly integrated. It's part of the plone ecosystem and we should make testing in that ecosystem as simple as possible. Tests for mockup still run outside of a plone environment in travis.

jcbrand:
> All @bloodbare is saying is that mockup is also a plone project and tightly integrated.

If the goal is to do (or at least allow) dev outside of Plone, then why is it tightly integrated into Plone?

Tight integration into Plone appears to me to be orthogonal to the idea of being able to do mockup development outside of Plone. This pull request looks to me like it was meant to address that.

vangheem:
mockup is tightly integrated in Plone. Plone is not tightly integrated into mockup.

vangheem:
And yes, this pull request did address being able to install mockup as a bower package, which we want, it just did it in a way(with plone.mockup), that we disagree with for the overall Plone project.

jcbrand:
> And yes, this pull request did address being able to install mockup as a bower package, which we want, it just did it in a way(with plone.mockup), that we disagree with for the overall Plone project.

Can you please explain what's wrong with this approach? It's still not clear to me why it's fine that mockup-core can be a standalone pure JS library but the mockup patterns can't be.

What's the current development and testing workflow and how would it change for the worse if we have separate mockup and mockup.plone?

vangheem:
Why does it matter if we just keep one package mockup? Why are you so keen on splitting it up?

jcbrand:
@vangheem I didn't make this pull request nor did ask for it to be made.

I assume @thet has his reasons but until he mentions them I can only speculate.
I gave a thumbs up because I'm pro making mockup available on bower and in general I think it's preferable to not have a bunch of python in a JS package.

However, I'm trying to understand the pros and cons of the different approaches. That's why I asked you those questions and would appreciate having them answered.

vangheem:
I don't feel like arguing about it. The cons have been discussed on the plone developer list already. Basically, git submodules are hard to deal with, johannes jumped the gun on the change without discussing and there is a way to make mockup bower installable again without doing this. Johannes is also in agreement that it should be reverted.

jcbrand:
Ah, I missed that discussion. Will read it now.

thet:
reverted!

i reset to the commit before the merge and cherry-picked the one after, so that the commit history isn't again reset at the point where the files were moved.

i'll use this branch for project-specific bower package, where i want to depend on mockup until we come up with a better solution to mockup's bower installability.

thet:
These are the related pull requests on other repositories:
https://github.com/plone/Products.CMFPlone/pull/393
https://github.com/plone/plone.app.widgets/pull/89

The related commit on buildout.coredev 5.0:
https://github.com/plone/buildout.coredev/commit/c1f80c942cdc8545204ec9ecd4817a86e3369685

And the related plone.mockup package:
https://github.com/plone/plone.mockup

Let's keep these branches/repos until we have a better solution to the bower package problem.

PR #468, opened 2015-03-26
WIP: Bootstrap tooltips

obct537:
Replaced the existing tooltips pattern with one that includes the bootstrap tooltip. This fixes an issue where the $.fn.tooltip namespace is overwritten by the jquery tools tooltip, which breaks the buttons in the filemanager pattern.


vangheem:
for those watching, I've asked @obct537 to pull out the tooltip code from bootstrap and move here in a way that isn't use $.fn.tooltip. The reason is the we are getting conflicts with bootstrap tooltip and jquery tools tooltip for some of our patterns in plone--nasty. So we're just trying to avoid using $.fn._ wherever we can.

thet:
That explains, why i got issues at a current project with Bootstrap tooltips.

Does it help, if I pull-out the jquerytools tooltip code into separate files, which can be deactivated in the resource registry? We already forked jquerytools code anyways, so we can fix that there too.

vangheem:
jquery tools tooltips aren't being used anywhere right? I would be okay with that. It is probably still useful to put bootstrap tooltips in this pattern so we don't have 2 implementations being used.

thet:
+1 for using bootstrap tooltip in this pattern. i'm going to rip out tooltips from jquerytools, but i'll put it in a file, which is included by default for plone 4.

thet:
tooltip separated in plone.app.jquerytools https://github.com/plone/plone.app.jquerytools/commit/c2ebd5d19731c1797d3392491961eec54c588765

we do get this package out of plone 5, do we? it sucks.

vangheem:
mockup pulls in jquery tools from jquery.recurrenceinput.js--could you update that repository's jquery tools and then update the mockup bower.json version? https://github.com/plone/mockup/blob/master/bower.json#L25

PR #470, opened 2015-04-02
Change index references from Type to portal_type

thet:
Change all index references from Type to portal_type. E.g. the TinyMCE configuration option containsobjects expects portal_type values, not Type.

I had an issue with having 'Document' declared as folderish via an tinymce.xml 'containsobjects' entry, but the TinyMCE pattern wasn't respecting it (even with some necessary changes in plone.app.widgets.utils, for which i'll create a separate pull-request). It was due to mockup querying for the Type index, which is the title of the content type. In my case, the title is not Document, but Page.

IMO, we should remove the Type index at all and only use portal_type, but this is another discussion (or PLIP).

See https://github.com/plone/Products.TinyMCE/blob/master/Products/TinyMCE/profiles/default/tinymce.xml#L15 for tinymce declaring "Document" as "linkable" type, which is a proof that the TinyMCE config expects the content type id / portal_type and not the content types' title.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,15 @@
 Changelog
 =========
 
+2.0.3 (unreleased)
+------------------
+
+- Change all index references from ``Type`` to ``portal_type``. E.g. the
+  TinyMCE configuration option ``containsobjects`` expects portal_type values,
+  not Type.
+  [thet]
+
+
 2.0.2 (2015-04-01)
 ------------------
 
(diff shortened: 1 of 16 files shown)


vangheem:
Yes, I understand how this would be necessary for any querying and usage of types. Only thing I wonder is if there are cases we'd want to display "Type" metadata to the user instead of "portal_type" and we'd want both. In any case, I'm going to merge and we can modify attribute usage and templates if we see a problem.

thet:
Thanks for the merge.
You brought up a good point. We might also need translations of type names. E.g. the ReallyUserFriendlyTypesVocabulary (or so) returns vocabulary items with the translated "Type" and values, which are "portal_type" values. Maybe we can do something like that, if we come across the necessity to do so.

PR #478, opened 2015-04-09
fix param passed when setting default page from action menu in structure

davisagli:
refs plone/Products.CMFPlone#424


jcbrand:
@davisagli: Attributes on Backbone models should be fetched with the ``get`` accessor.

So the code should rather be ``'id': this.model.get('id')``.

jcbrand:
I just double checked that, and I see that they ask that ``set`` be used for modifying attributes and don't specifically say that ``get`` must be used.

Nevertheless, they refer to the ``attributes`` property as an "internal hash", so I think it's a good idea to avoid using it directly and rather sticking to ``get`` and ``set``.

http://backbonejs.org/#Model-attributes

Issue #476, opened 2015-04-08
Text editor fullscreen mode


ebrehault:
In the theme mapper, the user should be able to switch the text editor to fullscreen (just like the inspectors already do).

ebrehault:
Fixed by #498

PR #498, opened 2015-05-20
Thememapper fullscreen

obct537:
Added a button to the thememapper that shows the Ace editor window in fullscreen, as requested in #476. Also fixed a minor issue causing Ace to resize incorrectly when its container size was changed.


vangheem:
fwiw, take a look at how some of the other patterns do javascript translations.

"title: 'Fullscreen'" would change to "title: _t('Fullscreen')" so we can get translations of these things in JS.

ebrehault:
Excellent, so I considered #476 as fixed.

PR #500, opened 2015-05-21
Thet twbs

thet:
- Update to jQuery 1.11.3 and moment 2.10.3.
- Use ``windows.alert`` and ``window.confirm`` instead globals.
- Remove bootstrap-tooltip from requirejs config, as we have our own.
- Define more export variables for Bootstrap plugins.


vangheem:
lucky pull request #500!

thet:
:)

thet:
these were really minor changes - i wanted to get this cleaned up before debugging a problem with bootstrap-dropdown in a customer project.
conclusion: you cannot include bootstrap-dropdown in mockup AND in your theme, e.g. that leads to both dropdown code paths being executed and the dropdown not working. i removed bootstrap-dropdown from my theme.

PR #504, opened 2015-05-28
Add test-jenkins

gforcada:
This target runs mockup tests and saves the results in an XML file
so that jenkins can show them with its publishers.

Depends on https://github.com/plone/mockup-core/pull/29


gforcada:
@vangheem @thet @frapell can we get this merged so we can start testing mockup on jenkins? Right now only on PhantomJS but later on at least Firefox and maybe Chrome and via SauceLabs (@datakurre help appreciated :) IE/Safari/mobile and so on...

vangheem:
Oops. I apologize. I missed this one.

gforcada:
@vangheem thanks!

PR #505, opened 2015-05-29
Allow to provide a sort_on and sort_order attributes for the QueryHelper

frapell:
This allows to specify an order in which results for the tree will be returned.

I'm creating a PR instead of directly commiting, because this will affect the related items widget, where returned items will be first folders, and then items.

In my opinion, we should also sort those folders and items alphabetically, but that should be done in the vocabulary, not here.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,9 @@ Changelog
 2.0.4 (Unreleased)
 ------------------
 
+- Allow to provide a sort_on and sort_order attributes for the QueryHelper
+  [frapell]
+
 - handle errors better with the modal pattern
   [vangheem]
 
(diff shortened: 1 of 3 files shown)


vangheem:
Can you also modify the defaults then? I don't think

sort_on: 'is_folderish'

is a good default.

vangheem:
Otherwise, looks fine. Merge whenever you like.

frapell:
Well, I think is a good idea to get folders first and then items... I mean, that's what most people expect when navigating through files and folders, right?

frapell:
Although seems I need to fix some tests before ;)

vangheem:
Oh right, that does make sense.

frapell:
@vangheem How can I trigger the tests to run again?

vangheem:
Sign into travis and restart the build. Otherwise, on commit, it should auto restart.

If tests are passing locally, that's fine too.

frapell:
@vangheem well, I restarted them but for some reason is not including the changes from commit 817cbe7 so it is still failing.

Tests are passing locally though, so I will merge it

vangheem:
Yah... you can't trust travis half the time...

frapell:
Merged in fa9779cc9df885575e8a55fc3288fcc7a9d5b66b

PR #511, opened 2015-06-23
Rulebuilder

obct537:
Converted Plone 4.3 diazo rule builder to work in thememapper pattern. Moved the RuleBuilder class into it's own file, separate from the Thememapper, and created a new popover view to display it. 

Also @vangheem *FINALLY* fixed the infamous "2000 error per minute" scrolling bug in the Ace editor. 

Added in a basic testing setup for the thememapper.


vangheem:
Great. Thanks @obct537 !

PR #517, opened 2015-07-08
fix select2 widget use of allowNewItems

vangheem:


pbauer:
Looks good and with the changes it works for me. But the build is failing: https://travis-ci.org/plone/mockup/jobs/70063569

vangheem:
alright, test fix pushed.

PR #518, opened 2015-07-08
add feature detection support

vangheem:


vangheem:
@obct537 I didn't mean to merge this. Oops. Can you review and comment on what you think about it?

obct537:
Seems fine to me. The comment you just added is probably a good idea too :)

PR #520, opened 2015-07-13
show quick edit/view for items in structure row

vangheem:
This is to address https://github.com/plone/Products.CMFPlone/issues/524

Here is a screenshot:
screen shot 2015-07-13 at 6 17 38 am

Thoughts?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,9 @@ Changelog
 2.0.5 (unreleased)
 ------------------
 
+- show quick edit/view for items in structure row
+  [vangheem]
+
 - upgrade mockup-core to 2.1.10
   [vangheem]
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,7 @@ Changelog
 2.0.5 (unreleased)
 ------------------
 
-- show quick edit/view for items in structure row
+- show quick view for items in structure row
   [vangheem]
 
 - upgrade mockup-core to 2.1.10
--- a/mockup/patterns/structure/js/views/tablerow.js
+++ b/mockup/patterns/structure/js/views/tablerow.js
@@ -13,7 +13,7 @@ define([
     template: _.template(TableRowTemplate),
     events: {
       'change input': 'itemSelected',
-      'click td.title a': 'itemClicked'
+      'click td.title a.manage': 'itemClicked'
     },
     initialize: function(options) {
       this.options = options;
--- a/mockup/patterns/structure/js/views/tablerow.js
+++ b/mockup/patterns/structure/js/views/tablerow.js
@@ -3,8 +3,9 @@ define([
   'underscore',
   'backbone',
   'mockup-patterns-structure-url/js/views/actionmenu',
-  'text!mockup-patterns-structure-url/templates/tablerow.xml'
-], function($, _, Backbone, ActionMenu, TableRowTemplate) {
+  'text!mockup-patterns-structure-url/templates/tablerow.xml',
+  'mockup-utils'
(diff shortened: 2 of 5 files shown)


jensens:
At least the UI/UX team should have a look here (who to ping?)

agitator:
@plone/ux-team what do you think about about @vangheem proposal?

I'd say the "eye" icon would be enough to solve that problem and to not add complexity

davilima6:
My suggestion:

folder-contents-actions

vangheem:
@davilima6 +1

davilima6:
Btw any special reason for dropping the sorting arrows which stayed next to each column's title in Plone 4? I think that's way more intuitive usability-wise and would also help reducing the amount of buttons in that navbar (which now takes 3 (!) rows in screens smaller than 1200px).

vangheem:
If I understand you correctly, rearrange is not meant to be a replacement for changing the display sort order. It permanently rearranges the items in the folder. It needs to be a button.

If you want to change the display sort on the items, that is done with the Query button.

davilima6:
Thanks, only now I understand. Rearranging is a very powerful new feature, but also potentially dangerous imho since most sites base their navigation on folder position. My concern is similar to what currently happens with the Display menu: too often editors misconfigure their homepages and sections because that option is too visible and offers no contextual help so people just click away to test it. Maybe we should offer some warning that operation cannot be undone (just as when deleting an item)?

vangheem:
Yes, this has been mentioned by others too. I think a warning confirmation prompt when you are on site root will suffice. Please create a ticket.

If you implement the arrows on the columns, please just make sure it is updating the query widget so those don't get out of sync.

davilima6:
Opened: https://github.com/plone/mockup/issues/526

djay:
I'm not sure the word workflow is used much elsewhere. it should be the same as the toolbar right?
I do agree rearrange is still likely to be mistaken for sorting the display. but other than a warning I'm not sure how to hide it. its really a power feature that is not going to be used much so it shouldn't be so prominent.

djay:
BTW the quick edit buttons being closer to the left looks good to me. Maybe
a little small. Have to see if its fiddly to push.

davilima6:
Did you really mean the workflow word? I imagined you meant Rearrange but that's not on the toolbar, is it?

I tried the Actions column on the left but it didn't look good as it related too much with the checkboxes. Since those are meant to be used on _batch_ actions, IMO having those markers on the left and individual item controls on the right resolved things clearer.

It will be less mistaken if the old arrows are there (instead of hidden inside a

davilima6:
+1 on making Rearrange less prominent and suggest "hiding" it under "Actions on current folder", labeled "Rearrange items", just after "Select all contained items".

Btw I also suggest adopting `glyphicon-options-horizontal` there to match the one in `Actions` column (where users can also cut'n'copy an item). Because it's "more" not "configuration".

vangheem:
no, rearrange won't work well on actions dropdown. A popover on a dropdown won't work. A warning message will suffice.

agitator:
+1 for the proposal of @davilima6 
and please stick to the meaning of the issue title

djay:
I think the actions are best just to the right of the title. If you are navigating around and doing actions on items then don't want to be going all the way over the to right. I don't mind @davilima6 suggestion if it was in the col to the right of title but @vangheem takes less space.

vangheem:
We're all over the place here. I'm merging only to have one icon--link to the view of the item. It's less impactful and I think is a fine compromise here.

agitator:
changed "workflow" to "state"

PR #524, opened 2015-07-14
fix mockup pattern querystring, for multible items in multiselects

MrTango:
… like types and tags


vangheem:
test failure is just travis/sauce not working.

vangheem:
Thanks!

PR #529, opened 2015-07-16
fix pickadate default timezone to work even if the default isn't the …

braytonosg:
…last timezone in the list


vangheem:
can you rebase please?

PR #534, opened 2015-07-17
fix pickadate default timezone to work even if the default isn't the …

braytonosg:
…last timezone in the list

Trying this again, as my first attempt to rebase failed


vangheem:
Tests are fine. It is just a problem with sauce starting...

PR #539, opened 2015-07-23
be able to translate tablerow title

vangheem:


pbauer:
Is the test-failure some real or only travis being unreliable again?

vangheem:
Just another sauce connection error... Just ignore. I should just disable attempting to run the sauce tests..

PR #540, opened 2015-07-26
remove unused tablesorter pattern

vangheem:


vangheem:
@thet @bloodbare @frapell anyone, can we remove this pattern? It's not used in core anywhere and it doesn't seem very useful unless you have all the contents in one table.

bloodbare:
+1 to remove it

hvelarde:
:-1: this is something that is pretty useful when creating custom views and we have such functionality in Plone 4 and before.

jensens:
really? it did not worked batched. maybe it better lives as an addon then. or implement something working with batching. I used jquery datatables with its json features a lot in past.

hvelarde:
FYI @agnogueira

domruf:
BTW [mockup-training](https://mockup-training.readthedocs.org/en/latest/including_from_mockup.html#pick-the-pattern-to-use) uses Table Sorter in its examples.

davilima6:
-1 for removing this one. I think this is [old and gold](http://pydanny.blogspot.com.br/2008/02/sorting-by-dates-in-plone-tables_07.html) Plone builtin that [many content editors](http://www.sixfeetup.com/blog/utilize-available-javascript-in-plone-without-knowing-javascript) got used [to expect](https://plone.org/documentation/kb-old/making-tables-sortable). And now that it's a pattern it can be expected to be made more configurable (eg: defaults for asc/desc or date formatting). Besides, only reason it's not in core right now is it hasn't been reimplemented in new folder contents - in the same manner multiple paste of copied items wasn't available until recently or that some configlets functionality were temporarily lost due to z3c.form migration.

+1 for both this being left available to content editors and to use it in new folder contents, even if we still don't have ["View all items in one page" (#735)](https://github.com/plone/Products.CMFPlone/issues/735) implemented.

vangheem:
No, this has nothing to do with folder contents. This pattern has zero relevance to folder contents.

This pattern is NOT used anywhere in core, is not maintained and does not working with paginated result sets. I find it very hard to understand why it should be supported in core.

davilima6:
I'm sorry if I'm mistaken but isn't this the pattern that allows the [reinsertion of arrows in folder contents headers](https://github.com/plone/mockup/pull/520#issuecomment-121605048)?

vangheem:
No. This is a pattern that can be used with a table on some addon's custom view.

The feature you are talking about with folder contents has nothing to do with this pattern.

frisi:
personally i think this should not have been removed. @davilima6 arguments are valid (except the one for using it in folder_contents) and as @domruf points out it's even used in the documentation.
tablesorter.js used to be a nice feature that also allowed content editors to provide sortable tables (by assigning the corresponding class to the table in tinymce)

until right now I thought that all patterns listed on http://plone.github.io/mockup/dev/#pattern/ do at least have registered resources shipped by plone so I can include them in my custom require statement as [shown in the docs](https://github.com/plone/mockup-training/blob/master/including_from_mockup.rst)

PR #549, opened 2015-08-06
Thememapper updates

obct537:
Fixes/adds several features to the thememapper:
- Let's users specify a filename when building LESS into CSS
- Adds button to reload a file from server
- "Build CSS" button is now disabled, instead of hidden when editing a non-less file, to maintain a      consistent button layout
- Pasting text into the Ace editor will now mark the file as "modified" properly
- The filemanager tree now stays open after a add new/delete/upload/rename
- Renaming a file to a previously-but-no-longer-existent filename will now correctly load the new version of the file
- Slightly changed the highlighting scheme in the thememapper tree to make folder selection more intuitive
- Changed styling of the toolbar buttons to be both more consistent with Plone, and to allow for more buttons on screen

There's a lot going on here, I probably should have broken this into more than 1 pull request ;)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,23 @@ Changelog
 
 2.1.0 (unreleased)
 ------------------
+- add "Save As" option in less builder
+  [obct537]
+
+- add Refresh button to filemanager
+  [obct537]
+
+- filemanager tree now remains open after add/delete/rename/upload
+  [obct537]
+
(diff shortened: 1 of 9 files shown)


obct537:
For some reason CHANGES.rst is showing up kinda funny in the diff, but that should be up to date.

ebrehault:
Once again, great job on the thememapper, thank you @obct537 !

obct537:
Appreciate it! @ebrehault

PR #554, opened 2015-08-11
Fix the regular expression

ale-rt:
Regular expression was failing when the data-attribute is not the last attribute in the tag, see:
 - https://regex101.com/r/pI4kL9/1

I wanted to use jQuery instead of regexp bacause:
 - we have it
 - it is less error prone
 - it is more clear

but it turns out it is too complicated to be cross browser.


vangheem:
You need to look at the response, not the current DOM.

ale-rt:
Yes, I agree... will fix this when I am less tired

ale-rt:
@jensens @vangheem IMHO this is already mergeable

ale-rt:
Thanks @vangheem

PR #558, opened 2015-08-21
Manifest

obct537:
The lessbuilder will now guess filenames based on a theme's manifest.cfg file.

Popovers will now close when the filemanager is switched to another file.

Updated some of the buttons in the filemanager to make them more consistent

### DIFF
--- a/mockup/patterns/thememapper/js/lessbuilderview.js
+++ b/mockup/patterns/thememapper/js/lessbuilderview.js
@@ -13,8 +13,8 @@ define([
       '<div class="buttonBox">' +
         '<label for="lessFileName">Save as:</label>' +
         '<input id="lessFileName" type="text" placeholder="filename" />' +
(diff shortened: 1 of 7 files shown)


mockup/patterns/thememapper/js/lessbuilderview.js
```diff
@@ -52,10 +52,22 @@ define([
...
     setFilename: function() {
         var self = this;
+
         if( self.app.lessPaths['save'] === undefined ) {
             return;
         }
-        var f = self.app.lessPaths['save'];
+
+        var filePath = self.app.lessPaths['less'];
+        var devPath = self.app.devPath[0];
```
vangheem: Will this be okay if no dev path is set in the manifest?

obct537: Yep, it will just change the current filename from .less to .css

PR #568, opened 2015-09-08
tinyMCE inline editing

thet:
Add option to use tinyMCE inline on a contenteditable div. The pattern creates
the contenteditable div from the textarea, copies it's content to it and
handles saving changed data back to the textarea on form submit.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,10 +1,13 @@
 Changelog
 =========
 
-2.0.7 (Unreleased)
+2.0.8 (Unreleased)
 ------------------
 
-- No changes yet.
+- Add option to use tinyMCE inline on a contenteditable div. The pattern
+  creates the contenteditable div from the textarea, copies it's content to it
+  and handles saving changed data back to the textarea on form submit.
+  [thet]
 
 
 2.0.7 (2015-09-07)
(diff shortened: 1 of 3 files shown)


vangheem:
Isn't this a little backwards? Wouldn't you want to just edit an existing div, not create a new div from a textarea?

thet:
that was my first approach. but then you'd need to insert a div for rich text fields. we would need to support that somewhere in plone.app.z3cform or plone.app.textfield and we would need to write javascript code to copy the contents of the div into a input element on form submit.
i think this approach is much simpler and should even work with javascript disabled (the "div instead of textarea" method wouldn't).
don't you think so?

vangheem:
I understand. Should it be another pattern that wraps this pattern then? I'm not totally against it, I just wasn't sure if this was the right place for it.

thet:
Another pattern would be possible.
But I was thinking of eventually exposing an "Inline Editing" option in the tinyMCE controlpanel, so integrators can easily switch between iframe and inline editing. Would this switch be possible with two different patterns?
The reason why someone might want to switch between iframe and inline editing is, that inline allows to use the same styles as the rest of the page, making the editing content looking the same as it would when just displaying.

thet:
Besides of that, I found an issue when switching to another mimetype with the textareamimetypeselector pattern.

vangheem:
Ah, I see. I misunderstood how you were using this.

thet:
tnx for merging! currently i'm struggling with fixing the issue with the textarea mimetype selector pattern.

PR #571, opened 2015-09-09
Raise pat-modal z-index

thet:
Raise pat-modal ``z-index`` to make sure, it's displayed above anything
else, e.g. TinyMCE's toolbar in inline editing mode.


thet:
this breaks tinymce dropdown menus in overlays. see: https://github.com/plone/mockup/issues/573

i'll revert.

PR #580, opened 2015-09-16
Inline tinyMCE / mosaic fix

thet:
Fix inline tinyMCE to work together with mosaic. The inline option must now be passed to the patterns option object instead to the patterns tiny options object.

### DIFF
--- a/mockup/patterns/tinymce/pattern.js
+++ b/mockup/patterns/tinymce/pattern.js
@@ -187,7 +187,7 @@ define([
                  'bullist numlist outdent indent | ' +
                  'unlink plonelink ploneimage',
         //'autoresize_max_height': 900,
-        'height': 400,
+        'height': 400
       },
       inline: false
     },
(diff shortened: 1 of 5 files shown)


datakurre:
@thet Thanks a lot! Trying this out soon.

datakurre:
@thet I can confirm that this fixes Mosaic on Plone 5.0rc2. Thanks a lot.

thet:
... if this travis build error has something to do with these changes...?

mockup/patterns/tinymce/pattern.js
```diff
@@ -187,8 +188,8 @@ define([
                  'unlink plonelink ploneimage',
         //'autoresize_max_height': 900,
         'height': 400,
```
datakurre: @thet Could that single comma break something?

datakurre:
@thet Green now. Some might have been fixed by that comma. Some (firefox at least) may be just flickering.

thet:
Great! Didn't know, that Javascript engines are so picky about unnecessary commas in object literals. Thought, only JSHint/JSLint is.
I'm merging.

PR #585, opened 2015-09-19
Remove mockup-core

jcbrand:
This pull request removes mockup-core completely as a dependency of mockup.

Changes made:
- The Base pattern from mockup-core is now completely replaced with the Base pattern in Patterns/src/core/base.js
- Moved the mockup-core parser to Patterns/src/core/mockup-parser.
- Changed the mockup patterns to explicitly declare that they use the mockup parser. The reason for this is that in the long term we want patterns to use the Patternslib parser and deprecate the mockup parser. I didn't yet find a way to merge/combine the two.
- ~~Unfortunately Mockup depends on an older version of Lodash which is incompatible with the newer Underscore version which Patterns depends on. Using one or the other causes test failures and problems, so for now we are loading both.~~
- The mockup-core docs JS and less files and its Grunt config have all been moved into mockup.
- Updated bower.json to depend on the master of Patterns (the changes required are not yet in a released version) and include bower dependencies from mockup-core (react.js etc.).

If you set up your environment with this branch mockup-core won't be pulled in at all.

All tests pass locally on my machine, hopefully will on Travis as well.

@frapell I notice you added a method `_recursiveExtend` (as mentioned [here](https://github.com/plone/mockup-core/issues/31)). Initially I added this to the Patterns base class, but adding that method caused tests to fail for the pickadate and tinymce patterns, so I removed it again. I'm not sure needs to be done there. Currently mockup doesn't depend on the verion of mockup-core where you added that method.


vangheem:
Very nice!

Weird though. I thought we were already using patternslib parsers. The parser is the same as the scanner right? That's the only part of this PR I don't like.

jcbrand:
Since the Innsbruck sprint the Patternslib registry and scanner were being used, but I was not yet able to replace the Mockup parser.

The Mockup parser accepts anything that looks like declarative **data-*** configuration, while the Patternslib parser must first be told what arguments to accept (via the **addArgument** method) and will ignore anything else.

So getting rid of the Mockup  parser would mean updating all the Mockup patterns to explicitly make `parser.addArgument` calls for all the arguments that those patterns accept. Doable, but a lot of work.

Now that all Mockup patterns explicitly declare that they use the Mockup parser we can update them one at a time to use the Patternslib parser. 

So concerning the parser, the only change now was to move the Mockup parser (which was still used) out of mockup-core and into Patternslib and to update the patterns to explicitly state that they use it.

vangheem:
I see.

One other thing. Since this is so late in the game, could we have a "mockup-patterns-base" requirejs include that is a shim with the new patternslib base class with the "mockup" parser set? I'm worried this will break a lot of things.

Also, docs and training will need to be updated.

\cc @pbauer @frapell @thet

jcbrand:
@vangheem Are there patterns outside of plone/mockup which use mockup-patterns-base?

thet:
@jcbrand - awesome!

In a client project, I use a pattern for client specific/theme related
code. I'd have to update the pattern, which isn't too hard for me.
There is a new collective.isotope package, which is uses mockup. And
there is the outdated mockup-extra. That's what I recall.

Cheers,
  Johannes

jcbrand:
Thanks @thet :)

I'll have to think a bit on how best to create such a shim. I won't be able to do this until later in the week. It's now 00:52 here and tomorrow I'm away. Going to bed now.

thet:
If it's only these three projects, it won't be worth the effort. But I guess, there will be more. Maybe a major version change justifies such a big change?

jcbrand:
There's unfortunately some weird heisenbug concerning TinyMCE which causes tests to fail intermittently.

Have any of you seen this before:
```
        TypeError: 'undefined' is not an object (evaluating 'doc.createRange') (/home/jc/buildout.coredev/src/mockup/mockup/bower_components/tinymce-builded/js/tinymce/tinymce.js
```

Googling shows that other people have had the same problem, and likely has to do with timing and DOM initialization. Haven't found a solution though :(

Happens about once every five test runs locally.

vangheem:
Yeah, I've been ignoring those random failures :)

frapell:
@jcbrand what are the tests that fail for pickadate and tinymce? I think the test from https://github.com/plone/mockup-core/blob/master/tests/pattern-test.js#L55 should cover all possible cases used in patterns, and if that test pass without my _recursiveExtend, then my guess is, it is not needed for patternslib...

As per projects using 'mockup-patterns-base', we are using it in a client's project, but I can either fix it, or just pin previous versions, so I wouldn't worry. I do agree that this type of change should go into a major release...

I think @thet and me would need a super fast crash course on this in order to be able to fit it into the training :)

jcbrand:
@frapell: When the Base pattern uses _recursiveExtend, then various tests in pickadate and tinymce fail, I think 6 tests in total. It looks like Pickadate doesn't get configured or instantiated correctly, probably because the options are different when _recursiveExtend is being used.

The current version of Mockup (pulled in from coredev 5.0 branch) doesn't use the version of mockup-core where you added that method, I'm not sure if there is a branch somewhere or how you tested everything together.

I'm happy to add that method again to the Patternslib Base pattern. I originally added it and then also added your tests which you wrote in [the mockup-core ticket](https://github.com/plone/mockup-core/commit/b093d5bf42aa95b91663e632b343543514af6108) to make sure everything worked as you intended. But then the pickadate and tinymce tests in Mockup failed, like I mentioned.

Here's where I added your method to Patternslib (I renamed it to **override**): https://github.com/Patternslib/Patterns/commit/48d94318b33897a6a19378d719fa708bd4cdff03

About a crash course, we can do a hangout this week if you are available, then we can discuss everything in detail.

jcbrand:
@vangheem I've been thinking a bit about the shim. I think this only makes sense if other patterns use the same require.js paths configuration file that Mockup uses (in `mockup/js/config.js`), which I don't think is the case.

Otherwise, if they have their own path configs, then those path configs will in any case have to be updated, because the path of the shim will be different (it'll be in mockup and not mockup-core).
In that case, we might as well just update those patterns to properly use pat-base.

However, if those other patterns have mockup-core as a bower dependency and have their config paths configured to point **mockup-patterns-base** to the mockup-core base pattern, then they should continue to work properly without any modification or shim.

petschki:
I also use mockup-patterns-base in some client projects. but not too hard to pin mockup(-core) or change/update my projects...

jcbrand:
@petschki Are your paths configured with the `mockup/js/config.js` file or do you have your own configuration file to configure the JS module paths?

petschki:
I usually extend from the `mockup/js/config.js` paths as described here https://github.com/plone/mockup-training/issues/4

jcbrand:
@petschki Thanks, that's a good reason to add a shim then.

jcbrand:
@vangheem I've added the shim.

vangheem:
Thanks. Looks good! The test failure is the random failure we've been getting.

jcbrand:
Thanks! Very happy about this change.

petschki:
Just tried it in a customer project. works like a charm without any changes. After simply changing the "mockup-patterns-base" to "pat-base" in my custom patterns everything is still working nicely! Thanks for this one...

jcbrand:
@petschki: Beware of the parser though! You need to also specify `parser: 'mockup'` when calling `Base.extend`, otherwise it will use the Patterns parser which is incompatible.

thet:
``make docs`` doesn't work anymore:
```
[thet@luft mockup]$ make docs
rm -Rf mockup/docs; mkdir mockup/docs; cp -R .git mockup/docs; cd mockup/docs; git checkout gh-pages;
Branch gh-pages set up to track remote branch gh-pages from origin.
Switched to a new branch 'gh-pages'
# if test ! -d mockup/docs; then git clone git://github.com/plone/mockup.git -b gh-pages mockup/docs; fi
rm -rf mockup/docs/dev
NODE_PATH=./node_modules ./node_modules/grunt-cli/bin/grunt bundle-docs   --gruntfile=mockup/Gruntfile.js
Running "requirejs:docs" (requirejs) task
>> Error: ENOENT: no such file or directory, open
>> '/home/thet-data/data/dev/plone/mockup/mockup/mockup/js/docs/app.js'
>> In module tree:
>>     mockup-bundles-docs
Warning: RequireJS failed. Use --force to continue.

Aborted due to warnings.
Makefile:57: recipe for target 'docs' failed
make: *** [docs] Error 6
```

thet:
Fixed: https://github.com/plone/mockup/commit/e7990d943174929150b1d2f9a89c63347346843b 
docs folder copied from mockup-core. You probably just forgot to adapt mockup's .gitignore, which was set to ignore all ``docs`` paths.

jcbrand:
Thanks @thet. I think you're right. I copied over the files but probably didn't notice that they're ignored by git.

thet:
Yes, it's hard to recognize.

PR #586, opened 2015-09-20
pat-moment: fall back to English

cillianderoiste:
Currently if you select a language which isn't supported by moment.js the last language in the list of available languages is used (Chinese). This PR uses English as the fallback instead.

See plone/Products.CMFPlone#1028

I'm having trouble adding a test for this:

      $('html').attr('lang', 'fr');
      var $el = $('<div class="pat-moment" data-pat-moment="format:relative">' + date + '</div>');
      registry.scan($el);
      var i18n = new I18n();
      console.log(i18n.currentLanguage);

Shows the language set to 'fr', but when I inspect pat-moment I see that new i18n()).currentLanguage is returning 'en-us'.

### DIFF
--- a/mockup/patterns/moment/pattern.js
+++ b/mockup/patterns/moment/pattern.js
@@ -105,7 +105,7 @@ define([
       if (!date) {
         date = $.trim($el.html());
       }
-      moment.locale((new i18n()).currentLanguage);
+      moment.locale([(new i18n()).currentLanguage, 'en']);
       date = moment(date);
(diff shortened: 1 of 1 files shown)


cillianderoiste:
I see that i18n gets configured in earlier tests, so running the locale test separately or resetting i18n (somehow) should help. The other problem is that we're not including the moment locale files in the tests at the moment, so moment.locale('fr') falls back to 'en'.

Issue #589, opened 2015-09-28
404 on pat-base.js


Gagaro:
Hello,

I have an issue when I have the plone-logged-in bundle in develop mode. I have a 404 on pat-base.js. I did a bower install to have all resources.

After bissecting, it seems to have be cause by the following commit: 8d616f966c016c7d26223d6c8d3d3aa70e8aaa56

Did I miss a step ?

Thanks.

jcbrand:
Looks like you don't have the path to pat-base configured.

Do you use custom patterns not included in the mockup egg?
If so, then those patterns probably have their own config file for specifying the paths of the JS dependencies. You then just need to add a path for pat-base as well.

Check in mockup/js/config.js for an example.

vangheem:
Products.CMFCore was never updated with the new base pattern I think.

vangheem:
Never mind, Plone never updated to latest mockup(which is a good thing at this point) so we're fine.

Gagaro:
I don't have any custom mockup that I know of. Is there a way to trace it ? The errors I have only point to require.js.

jcbrand:
From what @vangheem says, sounds like you need to add the path to Products.CMFCore. If you do a grep there for pattern names you'll probably quickly find the file where the paths are configured, then just add a path for pat-base as well.

If it works, please make a pull request.

Gagaro:
I tried adding it to Products.CMFCore but it still doesn't work.

Also `pat-base` is not in the `config.js` file I get in the browser. How is this file generated ?

jcbrand:
I can't say. I'm not familiar with how those things are done in Plone nowadays.

vangheem:
I'm sorry, I meant the latest Products.CMFPlone

We'd just need to add a registry entry for pat-base in https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/profiles/dependencies/registry.xml

Not sure this is your problem though.

jcbrand:
This is almost certainly a path issue.

Gagaro:
@vangheem I added the resource TTW (I'm not so sure how to apply this profile) and it seems to work. PR is there: https://github.com/plone/Products.CMFPlone/pull/1106

PR #590, opened 2015-09-28
feat: add option to select moment format in structure

Gagaro:
I would like to be able to choose the moment formatting in structures. As I am fairly new to mockup, I don't know if this is the best way to do it (I tried with html attributes but it seems to not working because of the Moment pattern being initialized in JS ?).

Please also note that this is not based on master because of #589.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,10 @@ Changelog
     in the `Base.extend` call.
     Refer to any of the core Mockup patterns for examples.
 
+- Add the ``momentFormat`` option to the ``structure`` pattern.
+  [Gagaro]
+
+
 2.0.12 (2015-09-20)
 -------------------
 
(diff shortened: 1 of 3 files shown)


jensens:
- changelog entry is missing
- your change the default from ``relative`` to ``calendar`` (no idea what implication it has), but it feels wrong.

Gagaro:
Good catch, I tested it and forgot to put back the relative value...

Changelog entry added.

Thanks for the review.

vangheem:
Did you re-test your change? I wonder if you changing the default is the only thing that made this work. I'm not sure ``self.options.app.momentFormat`` would get you the right option.

Gagaro:
I just retested it and it works fine. I find the app variable using a debugger as I wasn't sure how to get an option from a parent pattern. Please let me know if you know of a better solution.

vangheem:
Okay, good. I also didn't remember how those options were copied over :)

Could you rebase?

Gagaro:
Done

PR #596, opened 2015-10-16
fix: ensure we have all content for tree query in relateditems

Gagaro:
There is no changelog yet because it does not seem to be completely fixed and I'm not sure this is the better way to do it, so I'd like to discuss it first.

The issue I have is that for certain related widget (in my case, in the collection portlet), the `vocabularyUrl` option is set to `@@getSource`, which cause the results to be filtered by the field parameters (in my case, only `Collections`).

So it is impossible to browser the site using the related items widget in the Collection portlet form. I was able to kinda solve it with this fix but I'd like some input.

* Is `$('body').attr('data-portal-url')` the best way to get the portal url in a mockup ?
* I seem to only be able to browser a single level when using the right arrow (probably because it uses the results from the actual request), but I didn't find where it is done yet. If anyone have an idea, don't hesitate to comment :smile: .

Thanks.

### DIFF
--- a/mockup/patterns/relateditems/pattern.js
+++ b/mockup/patterns/relateditems/pattern.js
@@ -108,6 +108,9 @@ define([
       attributes: ['UID', 'Title', 'portal_type', 'path','getURL', 'getIcon','is_folderish','review_state'],
       dropdownCssClass: 'pattern-relateditems-dropdown',
       maximumSelectionSize: -1,
+      treeOptions: {
+        vocabularyUrl: null, // must be set to work
+      },
       resultTemplate: '' +
         '<div class="   pattern-relateditems-result  <% if (selected) { %>pattern-relateditems-active<% } %>">' +
         '  <a href="#" class=" pattern-relateditems-result-select <% if (selectable) { %>selectable<% } %>">' +
@@ -365,12 +368,13 @@ define([
       self.treeQuery = new utils.QueryHelper(
         $.extend(true, {}, self.options, {
           pattern: self,
+          vocabularyUrl: self.options.treeOptions.vocabularyUrl || self.options.vocabularyUrl,
           baseCriteria: [{
             i: 'is_folderish',
             o: 'plone.app.querystring.operation.selection.any',
             v: 'True'
           }]
-        })
+        }, self.options.treeOptions)
       );
 
       self.options.ajax = self.options.setupAjax.apply(self);
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -34,7 +34,11 @@ Fixes:
 
 - Fix `aria-hidden` attribute control problem on folder content panel
   [terapyon]
+
 - Trim links in tinymce before inserting them in the source.
+  [Gagaro]
+
+- Ensure we have all content for tree query in relateditems
   [Gagaro]
 
 2.1.2 (2016-01-08)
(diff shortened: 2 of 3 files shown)


vangheem:
Ahh, I see.

I think the related items pattern will need to take another option "treeOptions" that will have it's own vocabularyUrl set.

That being said, I think there are some UI problems with the tree widget part of this. It is only meant to traverse folders, not select an item. Selecting an item is still meant to be done with the select widget.

Gagaro:
Ok so I added the option. However I didn't find where the widgets are created and thus I couldn't change the option there (I let it as a default value).

vangheem:
See https://github.com/plone/plone.app.widgets/blob/master/plone/app/widgets/utils.py#L121

mockup/patterns/relateditems/pattern.js
```diff
@@ -350,6 +353,7 @@ define([
       self.treeQuery = new utils.QueryHelper(
         $.extend(true, {}, self.options, {
           pattern: self,
+          vocabularyUrl: self.options.treeOptions.vocabularyUrl,
```
vangheem: Can we use the whole treeOptions object here instead of just taking vocabularyUrl?

Gagaro: You mean using `treeOptions` to extends the option? Something like:

```Python
$.extend(true, {}, $.extend(true, {}, self.options, self.options.treeOptions), {
  pattern: self,
  baseCriteria: [{
    i: 'portal_type',
    o: 'plone.app.querystring.operation.list.contains',
    v: self.options.folderTypes
  }]
})
```

?

vangheem: Yes, I would do this though::

```JavaScript
$.extend(true, {}, self.options, {
  pattern: self,
  baseCriteria: [{
    i: 'portal_type',
    o: 'plone.app.querystring.operation.list.contains',
    v: self.options.folderTypes
  }]
}, self.options.treeOptions)
```

Gagaro: Right, thanks :+1: .

Gagaro:
So, I just went back to this PR. I added changelog, rebased the three PR and started a jenkins build.

http://jenkins.plone.org/job/pull-request-5.0/906/

Gagaro:
There is a test which errors in all jenkins builds. All other tests are green.

vangheem:
Jenkins doesn't really matter here until I create the compiled bundles in CMFPlone. Travis tests just have to pass.

mockup/patterns/relateditems/pattern.js
vangheem: So I'm still not convinced this will cause other problems. I want it to fallback onto the vocabularyUrl for the related items pattern.

I think there are cases doing it this way where vocabularyUrl will end up being null(since you default treeOptions.vocabularyUrl to null). Can you remove that default value?

mockup/patterns/relateditems/pattern.js
```diff
@@ -365,12 +368,13 @@ define([
       self.treeQuery = new utils.QueryHelper(
         $.extend(true, {}, self.options, {
           pattern: self,
+          vocabularyUrl: self.options.treeOptions.vocabularyUrl || self.options.vocabularyUrl,
```
Gagaro: Something like this?

vangheem: that'll work!

vangheem:
And now you need to rebase again. :(

Gagaro:
Done :ok_hand: .

vangheem:
Thank you!

Gagaro:
Thanks for the review :+1:

PR #598, opened 2015-10-18
Make formautofocus pattern work again.

### LINKED ISSUE #564: FormAutoFocus pattern seems to be broken
FormAutoFocus calls $().focusin() on the element but i think this is completely wrong as focusin() only defines a handler that is executed on a focusin Event[1] on this element. If i understand the FormAutoFocus correctly it should set the autofocus on the first input element of a form that has the .pat-formautofocus class.

[1] https://api.jquery.com/focusin/

timitos:
Fixes #564


jcbrand:
The failing test looks to be the tinymce timing issue again.

PR #600, opened 2015-10-21
fix: set widgets values for querystring pattern

### LINKED ISSUE #593: Why is ReferenceWidget default value not initialized ?
I am talking about this part specifically:

https://github.com/plone/mockup/blob/master/mockup/patterns/querystring/pattern.js#L253-L260

The issue I have is with Collection. Editing collection with path criterias will give empty inputs. I'm not sure what's the proper fix should be (changing them to StringWidget or adding the value there?).

So I'd like any input which could help me from people who know a bit more about this part.

Thanks! :smile:

Gagaro:
This is to fix #593.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -26,6 +26,9 @@ Changelog
 - Add the ``momentFormat`` option to the ``structure`` pattern.
   [Gagaro]
 
+- Set value for ``ReferenceWidget`` in querystring.
+  [Gagaro]
+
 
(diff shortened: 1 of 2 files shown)


Gagaro:
By the way, do I need to also provide the compiled version of mockup? And if I do, how should I do this?

vangheem:
I periodically create builds in the CMFPlone. Feel free to create a PR if it isn't done often enough.

PR #610, opened 2015-11-14
Fix flaky behavior of Resource Registries buttons 

davilima6:
(closes plone/Products.CMFPlone#1141)


vangheem:
Thanks! I'll make a new build

PR #615, opened 2015-12-08
Path criteria

obct537:
Changes how location queries are displayed in the querystring pattern.

@vangheem Please review this to death, when you get the chance.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -23,6 +23,9 @@ New:
   be taken care of with plone-modal-open class being applied to the body.
   This should fix issues with scrolling when this isn't properly cleared
   [vangheem]
+- Changed how the querystring pattern displays options for path-based queries,
+  to improve usability for less tech-savvy users.
+  [obct537]
 
 Fixes:
 
@@ -53,6 +56,8 @@ Fixes:
 ------------------
 
 New:
+- Fixed issue causing the querystring pattern to query multiple times per change
+  [obct537]
 
 - Added the ``momentFormat`` option to the ``structure`` pattern.
   [Gagaro]
--- a/mockup/patterns/querystring/pattern.js
+++ b/mockup/patterns/querystring/pattern.js
@@ -44,10 +44,11 @@ define([
   'pat-base',
   'mockup-patterns-select2',
   'mockup-patterns-pickadate',
+  'mockup-patterns-relateditems',
   'select2',
   'translate',
   'underscore'
-], function($, Base, Select2, PickADate, undefined, _t, _) {
+], function($, Base, Select2, PickADate, relatedItems, undefined, _t, _) {
   'use strict';
 
   var Criteria = function() { this.init.apply(this, arguments); };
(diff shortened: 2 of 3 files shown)


obct537:
https://travis-ci.org/plone/mockup/builds/95594464

vangheem:
yay!

frapell:
This is great! however, shouldn't the operation be defined in plone.app.querystring package?

vangheem:
It's using the same operator, just customizes the display/editing of it so the user doesn't have to think about depth param config, etc

vangheem:
@obct537 I started playing with this. There are a few issues that would be great to have addressed.

First thing, depth for "custom" version should be allowed to have "0" and "-1" values. "-1" would be an infinite depth. "0" is for selecting the object itself.

Next thing, the "advanced" selection I think should reintroduce the standard available operators that are normally overridden.

Then, I did notice this bug after opening, closing the location criteria:

screen shot 2015-12-10 at 11 05 23 am

obct537:
The repeated "Advanced" issue is bizarre....I've already encountered and "Fixed" that once before....I wonder why it's suddenly reappearing. 

Anyway, I'll take a look at all of this when I get a minute. Unfortunately, with the last week of school looming over me, that may not be for a few days.

obct537:
@vangheem In regards to the 0/-1 options for the Custom operator: Added a 0 option. There's already an "Any" option, that functions like the -1.....I changed the wording to "Unlimited" though, since "Depth: Any" doesn't really make sense (to me at least).

obct537:
Also, I'm looking at being able to switch back to the old operators. This is probably going to be fairly difficult though.

obct537:
@vangheem I believe I have all of your issues addressed now. Take a look, and more importantly, try to break it! :P

vangheem:
Looks great! I couldn't break it.

@frapell and any others? Not sure how much review we need on this.

I made a quick video so you can see how it works: https://www.dropbox.com/s/stdt1nqewpf775j/path-criteria.mov?dl=0

Comments please. It'd be great to get this merged!

fgrcon:
@video: wow, was missing this :-)

frapell:
This does look great, +1 for merging

I would create a ticket afterwards so we don't forget to add some tests around it

vangheem:
@obct537 Can you squash these commits?

If I don't get any disagreements on this, I'm going to be merging. It's very important that we fix the location widget to be more useful.

@obct537 also, can you add a couple robot tests to the Products.CMFPlone package?

thet:
Big +1 for this!

But i might have encountered some problems - probably related to upgrading old query strings.

Given the following: a Plone 4.3 site and mockup master and a Collection with the following query string:

```[{u'i': u'portal_type', u'o': u'plone.app.querystring.operation.selection.any', u'v': [u'Event']}, {u'i': u'review_state', u'o': u'plone.app.querystring.operation.selection.any', u'v': [u'mainportal']}, {u'i': u'path', u'o': u'plone.app.querystring.operation.string.absolutePath', u'v': u'/'}]```

On mockup master the querystring pattern renders like this:
screen shot 2015-12-23 at 14 22 27

On your branch ``path_criteria`` I get the following error:

```
TypeError: value.replace is not a function
        Criteria.prototype.createValue()    ++resource++amkumma-mockup.js:33225
        Criteria.prototype.createOperator() ++resource++amkumma-mockup.js:33034
        Criteria.prototype.init()           ++resource++amkumma-mockup.js:32915
        Criteria()                          ++resource++amkumma-mockup.js:32836
        QueryString<.createCriteria()       ++resource++amkumma-mockup.js:33512
        QueryString<._init/<()              ++resource++amkumma-mockup.js:33492
        .each()                             ++resource++amkumma-mockup.js:2516
        QueryString<._init()                ++resource++amkumma-mockup.js:33491
        QueryString<.init/<.success()       ++resource++amkumma-mockup.js:33448
        jQuery.Callbacks/fire()             ++resource++amkumma-mockup.js:5280
        jQuery.Callbacks/self.fireWith()    ++resource++amkumma-mockup.js:5392
        done()                              ++resource++amkumma-mockup.js:11446
        .send/callback()                    ++resource++amkumma-mockup.js:11850
```

And the querystring patterns renders like this:

screen shot 2015-12-23 at 14 29 17

thet:
The error raises on line 443 in the querystring pattern at this line: ``var trimmedValue = value.replace(/::[0-9]+/, '');``
 The value is: ``["Event"]``

obct537:
@thet Wasn't aware that value would be anything other than a string there. I'll add a check to prevent it from exploding on you :smile: 

@vangheem Testing is definitely on the to-do menu.

Issue #616, opened 2015-12-11
Question: would efforts going towards the decoupling the structure pattern from folder_contents be welcomed?


metatoaster:
The structure pattern looks quite useful for any kind of file/folder-like resources, however it makes assumptions on certain things, such as the availability of the paste button (which can be unavailable if the underlying view returns a json string without that button defined), and other Plone "folder_contents view"-centric ideas that make presenting a proxy file/folder view of non-Plone local resources (such as a view of some folder on the server, proxy view of git/mercurial repositories) rather difficult, and customization of action buttons isn't fully possible (i.e. paste button must be available even though the specific view implementation may not provide one).  I mostly want to know if this is within the scope of what plone mockup is wanting to achieve, so that no time is wasted on irrelevant goals.

djay:
I’d like to see the action buttons fully configurable like the only folder_contents was. ie coming from folder_buttons actions. At the moment its not possible to add your custom actions in there or remove any. Might be tricky to get the UI right to make it look ok with more actions but it should be possible.

vangheem:
@metatoaster Yes, we would welcome that. There was a branch a while ago that attempted to do that so we could reuse it to manage users and groups. I don't know what happened to it.

@djay yes, you can add custom buttons. See how workflow action is implemented for instance: https://github.com/plone/plone.app.content/blob/master/plone/app/content/browser/contents/workflow.py

metatoaster:
At least not overridable currently as how it is done now within `folder_contents`.  What I am trying to do is to create a completely new view that make use of structure to render some file/folder-like information sourced from a location external to the Plone site.  The view currently is loosely based on the `FolderContentsView` found in `plone.app.contents`, where the `get_actions` method is overridden, currently only returns a dummy field:

        def get_actions(self):
            return [{u'url': u'foo', u'title': u'Foo', u'id': u'foo', u'icon': u''}]

As is, clicking on the pagination button (or any interaction) as is will simply cause this be raised:

   TypeError: self.buttons.get(...) is undefined  [app.js:209:9](https://github.com/plone/mockup/blob/2.0.12/mockup/patterns/structure/js/views/app.js#L204)

Since app.js currently is hard-coded to assume that button exist, which in this case does not.  Naturally, providing a new action with the `...u'id': u'paste'...` will alleviate the issue.  First cut of the fix will have to check that button actually exists before trying to manipulate it, then later figure out how to include button-specific availability conditions to the `IStructureAction` implementations into the script.

[...]

PR #618, opened 2015-12-31
WIP: Making structure pattern more flexible to customization options

metatoaster:
This involves changes so that the actions don't make assumptions on the availability on certain key/value/attributes that are quite specific to `folder_contents`, grant full control over what columns are available and ensure that the settings can be made specific to one view only and not clobber others.

Done in response to issue #616.


mockup/patterns/structure/js/views/app.js
```diff
@@ -125,38 +125,62 @@ define([
...
-            }
-            var url = self.options.urlStructure.base + path + self.options.urlStructure.appended;
-            window.history.pushState(null, null, url);
-            $('body').trigger('structure-url-changed', path);
-          }else{
-            self.doNotPushState = false;
-          }
+        if (self.options.urlStructure) {
+          var url = self.options.urlStructure.base + path + self.options.urlStructure.appended;
+          window.history.pushState(null, null, url);
```
thet: You're not checking for availability of window.history.pushState here, like it was done before. Therefore loosing compatibility with IE < 10. (IIRC, we *wanted* to support IE>=8)

thet:
Tnx for the work on structure pattern!

metatoaster:
Thanks for doing the review and the merge.

If the `pushStateUrl` way of formatting the pushState are going to be supported (in 43b236d), the previous way with the `urlStructure` could probably be deprecated in the future.

PR #627, opened 2016-02-10
Structure selection well XSS fix

metatoaster:
The Title attribute for a given object could be used as an XSS attack vector. If the attacker can create new object or modify the title an existing object within a Plone instance, then convince her victim (or anyone with access) to use the folder_contents view and select the affected item, this would then populate the raw title text into the selection well without being sanitized, triggering the XSS attack.

The UID field is also vulnerable, but irrelevant as an attack vector for a typical Plone instance.

This pull request contains the test that demonstrates this problem, and then the fix to address this.

The status text set during copy/cut/paste is also vulnerable, however the i18n translation seems to sidestep this issue within the Plone instance context. That said, this pull request also correct this issue.

### DIFF
--- a/mockup/patterns/structure/templates/paging.xml
+++ b/mockup/patterns/structure/templates/paging.xml
@@ -11,7 +11,7 @@
     </li>
     <% _.each(pages, function(p){ %>
     <li class="<% if (currentPage == p) { %>active<% } %>">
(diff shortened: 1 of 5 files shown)


vangheem:
Thanks!

PR #628, opened 2016-02-11
Uncomment realteditems tests

metatoaster:
Uncommented tests for the relateditems pattern and corrected the issues in the test harness to make it work again.

PR #629, opened 2016-02-11
Relateditems pattern XSS fix

metatoaster:
The embedded template uses `<%= value %>` for everything which enables XSS attack.  Replaced all of this with `<%- value %>` and made a note for `<%= item %>` as that contains markup generated by the breadCrumbs template.  Requires pull request #628 as I need all existing test to run because overzealous cleansing broke breadCrumbs, heh.

As an aside, should better review be in place in the first place to actually not let tests be removed without proper explanation? Also maybe need some automated audit of the usage of the `<%= value %>` syntax as part of the tests, because I don't like XSS in my CMS.  Though this second part is a difficult requirement to address at the moment as I see it.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -26,6 +26,9 @@ Fixes:
 - Fix path widgets initialization in querystring pattern.
   [Gagaro]
 
+- Fix XSS vulnerability issues in structure and relateditem pattern.
+  [metatoaster]
+
 2.1.2 (2016-01-08)
 ------------------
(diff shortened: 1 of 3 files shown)


vangheem:
Thanks for this. I will say, if we have XSS in our catalog results for urls and paths, we're have other issues...

metatoaster:
Yes, found this out while doing my own QA on #618. This was not a fun one for me.

PR #634, opened 2016-02-17
Fix resource registry not allowing to go into development mode when  bundle is selected

vangheem:


vangheem:
fixes: https://github.com/plone/mockup/pull/634

jensens:
@vangheem  the fixes number is this ticket, please check if the right ticket was closed

PR #635, opened 2016-02-18
Fix `aria-hidden` attribute control problem on folder content panel

terapyon:
https://github.com/plone/Products.CMFPlone/issues/1194


vangheem:
Thanks!

PR #636, opened 2016-02-24
fix: treeQuery vocabularyUrl option in relateditems

Gagaro:
Fix for https://github.com/plone/mockup/pull/596

Adding `self.options.treeOptions` there overrides the `vocabularyUrl` set earlier and it doesn't fallback on `self.options.vocabularyUrl`.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -41,6 +41,9 @@ Fixes:
 - Ensure we have all content for tree query in relateditems
   [Gagaro]
 
+- Fix default value for treeVocabularyUrl in relateditems.
+  [Gagaro]
+
 2.1.2 (2016-01-08)
 ------------------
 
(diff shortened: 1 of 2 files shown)


vangheem:
perhaps instead of "treeOptions" we just provide a treeVocabularyUrl option?

Gagaro:
This doesn't change the fact that `self.options.treeOptions` should be removed there.

vangheem:
I know. Might as well provide one PR though right?

change the "self.options.treeOptions.vocabularyUrl || self.options.vocabularyUrl" line to "self.options.treeVocabularyUrl || self.options.vocabularyUrl"

Gagaro:
Sure, I'll need to update https://github.com/plone/plone.app.widgets/pull/114 and https://github.com/plone/plone.app.z3cform/pull/28 as well then.

vangheem:
Ahh, gotcha. forgot those packages had that config as well.

I suppose it might be good to leave it as is since it's already been introduced that way.

Gagaro:
I already updated everything. And it isn't merge yet, so it really is the best moment to do this change :) (and the resources are not compiled yet if I'm not mistaken).

vangheem:
Great. I'll re-run the jenkins tests and merge. Thank you.

PR #637, opened 2016-02-25
Use official TinyMCE bower dependency

thet:
Instead of the ``tinymce-builded`` bower dependency use the official ``tinymce`` one (refers to ``https://github.com/tinymce/tinymce-dist``).
This one is more recent and important bugfixes.
We still use ``tinymce-builded`` for the language files.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,11 @@ Changelog
 
 New:
 
+- Instead of the ``tinymce-bundled`` bower dependency use the official ``tinymce`` one.
+  This one is more recent and important bugfixes.
+  We still use ``tinymce-bundled`` for the language files.
+  [thet]
+
 - Fix resource registry not allowing to go into development mode when
   bundle is selected
   [vangheem]
--- a/bower.json
+++ b/bower.json
@@ -23,6 +23,7 @@
     "requirejs-text": "2.0.12",
     "selectivizr": "1.0.2",
     "select2": "3.5.1",
+    "tinymce": "4.3.4",
     "tinymce-builded": "4.3.0"
   },
   "devDependencies": {
--- a/mockup/js/config.js
+++ b/mockup/js/config.js
@@ -13,7 +13,8 @@
     'insertdatetime', 'layer', 'legacyoutput', 'link', 'lists', 'media',
     'nonbreaking', 'noneditable', 'pagebreak', 'paste', 'preview', 'print',
     'save', 'searchreplace', 'spellchecker', 'tabfocus', 'table', 'template',
-    'textcolor', 'textpattern', 'visualblocks', 'visualchars', 'wordcount'
+    'textcolor', 'textpattern', 'visualblocks', 'visualchars', 'wordcount',
+    'compat3x'
   ];
 
   var requirejsOptions = {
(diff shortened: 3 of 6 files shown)


thet:
Currently we're using TinyMCE 4.3.0 from tinymce-bundled. This change updates it to 4.3.4. For the changes see: https://github.com/tinymce/tinymce/blob/master/changelog.txt

thet:
Not sure, what to update for CMFPlone....
Probably Products/CMFPlone/static/bower.json, maybe also Products/CMFPlone/_scripts/_generate_gruntfile.py?
I see, there is still Products/CMFPlone/components/mockup-core, which is obsolete since mockup 2.0...

vangheem:
Any reason why you got rid of the "compat3x" plugin? This is required by the after the deadline plugin in Plone.

thet:
@vangheem ok - I'll re-add from tinymce-builded, a bit later today

thet:
@vangheem done

vangheem:
hhmmm, if we still need tinymce-builded for language files, why don't you just ask the author of tinymce-builded to update? Do we want to depend on 2 tinymce packages? Why doesn't the official one bundle language files?

jensens:
Working with official tinymce sounds more sane to me. Cant we add language file there on our own?

thet:
I'd prefer to add language files on our own. But they are located at transifex as .po files. I didn't find a way how to download them and build them automatically.

Asking the maintainer of tinymce-builded is an option, but we'd have to ask him quite often. I would prefer to be independend on the release cycles of tinymce-builded.

But tinymce-builded is also good for the compat3.x plugin, which I haven't seen in official repositories (maybe updating the "after the deadline" plugin makes compat3.x plugin obsolete? but the atd plugin for tinymce isn't very well documented - didn't find any version requirements...)

Another thing the tinymce-bundled is still used in this pull request is for the less files. Since we do not want to include the compiled CSS from TinyMCE (IIRC), we have to include selected LESS files. These are not available in tinymce-dist. They might be outdated in tinymce-bundled. So I thought I'd bring in a third tinymce dependency - the official unbuilded sources from github/tinymce/tinymce :)

Not sure how to proceed.
Build TinyMCE ourselves and not be dependend on official releases without language files or outdated inofficial releases?

thet:
The easiest thing would be to stick with ``tinymce_bundled``.
I asked for new release and contribution access https://github.com/artursmirnov/tinymce_builded/issues/6

thet:
... and I don't know why ``tinymce-dist`` doesn't include the language files. There was once an issue on https://github.com/tinymce/tinymce-dist but it looks they closed the issue tracker. IMO it makes no sense to create a tinymce-dist bundle without language files. However... That's probably why tinymce-builded exists.

thet:
Let's stick with ``tinymce-builded`` - I'll get contributor access and can then do updates on my own.

jensens:
great!

jensens:
may you document this somewhere in here?

frapell:
Yeah, I also found this when trying to pull from official tinymce, no idea why those guys do not include translation files.

FWIW, for automatic pulling from transifex, we use a python script http://pastie.org/10738442

http://docs.transifex.com/api/

PR #640, opened 2016-03-14
cleanup requirejs depenencies

thet:
... some code cleanup as part of getting to know, where backbone is used ...

i was careful to not remove any dependencies, wich are implicitly used and not used by a variable (pattern dependencies, which are needed by the generated output). tests pass, so i guess, this PR should be ok...

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -41,11 +41,15 @@ New:
 
 Fixes:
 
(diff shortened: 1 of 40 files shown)


vangheem:
Looks fine.

thet:
I will merge myself - after #618 to not complicate that merge more than necessary. currently #618 would not produce any merge conflicts.

PR #646, opened 2016-03-16
Added a new 'shadow-outline' pattern for mockup/patterns directory

khanstark:
'shadow-outline' pattern simply provides colorful boxed shadow to whichever element it is applied to, be it p,div or img.
It is auto-adjustable according to the height of the element. 
An index.html file is provided in order to view its implementation directly.
A dependency file is also added in the /src/ folder of the pattern.


khanstark:
I have updated `js/config.js `and `js/bundles/widget.js` files in the repo for implementation of the pattern.... 
to view it directly, there is a` pat.html` file in the root folder that shows the clear picture!!

vangheem:
@khanstark I'm not sure what you did but you will need to redo this PR. It contains over 250 commits and  485 file changes. I don't think that's what you're meaning to do. It is impossible for us to review.

khanstark:
I made it to the wrong branch thats why there were lots of commits...anyways, I reopened the PR with 10 commits and 7 file changes that will be reduced to 6 after your review.

PR #647, opened 2016-03-17
Added a new 'shadow-outline' pattern for mockup/patterns directory

khanstark:
Hope it works this time!


vangheem:
So why are you doing it this way? If a pattern is just a class applied to an element, wouldn't you just make the css instead of having a pattern set the css?

khanstark:
yeah we could absolutely do that.. but i thought i was making things easy for users by just making it a pattern that can be inserted directly.
Like i saw other patterns ,toggle, for instance that just made things bit easy or otherwise it was  just a matter of simple jquery.

vangheem:
Right, I'm just saying this sort of feature is better off in pure css, not JS perhaps.

khanstark:
alright...i agree with you.

PR #651, opened 2016-03-23
Structure pattern action items changes

thet:
Here is a pull request with changes to structure pattern for some more config options (action menu, icon size), code cleanup, better view url support:

- Allow definition of action menu items not only as dropdowns but also as buttons.
- Add ``openItem`` and ``editItem`` actions as buttons and remove the open icon from the title column.
- Open ``openItem`` links according to ``typeToViewAction`` instead of default with the ``/view`` postfix.
- Open ``editItem`` under ``/@@edit`` instead ``/edit``.
- Remove JS event handlers for externally opening simple URLs and use the href attribute instead.
- Add ``iconCSS`` option for action menus items to add icons.
- Add ``modal`` option for action menus items to allow links open in a modal.
- Add ``iconSize`` option to set the icon size if a item has an image.
- Use icons for all actionmenu entries.
- Use the tooltip pattern for all actionmenu buttons.
- Use pat-moment also for ``start``, ``end`` and ``last_comment_date`` columns.
- For columns with date fields, show an empty column if the date value is 'None'.
- Remove the checkbox and the actionmenu from the breadcrumbs bar for the current active folder to simplify the structure pattern.
  The actionmenu contained redundant actions (cut, copy, paste) and selecting the current folder is possible one level up.

The ``typeToViewAction`` configuration option uses sane defaults - the same as in a standard Plone installation. ``File``, ``Image`` and ``Blob`` are all mapped to ``/view``. ``plone.app.z3cform`` et al has to be changed to spit out these config options according to the ``plone.types_use_view_action_in_listings`` registry entries.

``typeToViewAction`` is an object in the form ``{'TYPE': 'VIEW_URL'}``, e.g. ``{'File': '/view'}``. Currently Plone uses only ``/view``. I thought making the view url configurable for each type makes the whole structure pattern a bit more generic. Maybe that's nonsense, comments welcome like for everything else.

Here are two screenshots showing the difference between old and new default action items.
Old:
folder_contents_old

New:
folder_contents_new

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -31,6 +31,7 @@ New:
   Open ``editItem`` under ``/@@edit`` instead ``/edit``.
   Remove JS event handlers for externally opening simple URLs and use the href attribute instead.
   Add ``iconCSS`` option for action menus items to add icons.
+  Use icons for all actionmenu entries.
   Add ``modal`` option for action menus items to allow links open in a modal.
   Add ``iconSize`` option to set the icon size if a item has an image.
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -20,6 +20,14 @@ Fixes:
 
 New:
 
+- Fix fakeserver ``relateditems-test.json`` response to return ISO dates for ``CreationDate``, ``ModificationDate`` and ``EffectiveDate``, as they really do in Plone.
+  This resolves a moment deprecation warning in structure examples.
+- Structure pattern: Allow definition of action menu items not only as dropdowns but also as buttons.
+  Add ``openItem`` and ``editItem`` actions as buttons and remove the open icon from the title column.
+  Open ``openItem`` links directly, omitting ``/view``.
+  Open ``editItem`` under ``/@@edit`` instead ``/edit``.
+  [thet]
+
 - set XML syntax coloring for .pt files in text editor
   [ebrehault]
 
@@ -42,11 +50,6 @@ New:
 
 - Alternative parameter/syntax for specification of the pushState url to
   be inline with the usage of ``{path}`` token in URL templates.
-  [metatoaster]
-
-- Structure can use the ``viewURL`` from a returned data item, alongside
-  with the previous default of simply appending ``/view`` to the
-  ``getURL`` attribute if this was not provided, for its view URL,
   [metatoaster]
 
 Fixes:
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -22,6 +22,7 @@ New:
   - Add ``iconSize`` option to set the icon size if a item has an image.
   - Use icons for all actionmenu entries.
   - Use the tooltip pattern for all actionmenu buttons.
+  - Use pat-moment also for ``start``, ``end`` and ``last_comment_date`` columns.
 
   [thet]
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -24,6 +24,8 @@ New:
   - Use the tooltip pattern for all actionmenu buttons.
   - Use pat-moment also for ``start``, ``end`` and ``last_comment_date`` columns.
   - For columns with date fields, show an empty column if the date value is 'None'.
+  - Remove the checkbox and the actionmenu from the breadcrumbs bar for the current active folder to simplify the structure pattern.
+    The actionmenu contained redundant actions (cut, copy, paste) and selecting the current folder is possible one level up.
 
   [thet]
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -30,6 +30,9 @@ New:
   Open ``openItem`` links according to ``typeToViewAction`` instead of default with the ``/view`` postfix.
   Open ``editItem`` under ``/@@edit`` instead ``/edit``.
   Remove JS event handlers for externally opening simple URLs and use the href attribute instead.
+  Add ``iconCSS`` option for action menus items to add icons.
+  Add ``linkCSS`` option for action menus items to allow additional CSS classes, e.g. to add a pattern like ``pat-modal``.
+  Add ``iconSize`` option to set the icon size if a item has an image.
   [thet]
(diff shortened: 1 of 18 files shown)


thet:
Missing: something like ``types_use_view_action_in_listings``, which adds a "/view" for ``File``, ``Image`` and ``Blob`` types.

metatoaster:
Could perhaps use a couple test to ensure the action groupings end up where they are supposed to be (i.e. category button vs. dropdown). Otherwise the change looks fine to me for the most part.

As for the the view action, I think it might be possible for the contextInfo view to export those information from the registry (i.e. `registry['plone.types_use_view_action_in_listings']`) which the value should then be accessible via `app.tableView.contextInfo` and the appropriate link can be triggered.

Thanks for cleaning up the list and turn that into objects (probably is something I should have done in the first place), and using that to generate the action urls for use in the action items so now we can start seeing URLs on the browser status bar for those things.

thet:
I think this is ready for review.

Tests pass - except tests in chrome and firefox on my local machine, but this might have some unknown system brokenness on my machine. See below, maybe someone has an idea?

@jensens mentioned it would be better, if the action menu also comes from portal actions. That would actually be great, but it's out of scope for this PR.

/cc @jensens @vangheem @metatoaster 

```
Chrome 50.0.2661 (Linux 0.0.0): Executed 57 of 63 (11 FAILED) (skipped 1) (0 secs / 49.845 secs)
    ✗ test navigate to folder push states - pushStateUrl
	Failed to execute 'pushState' on 'History': A history state object with URL 'http://localhost:8081/traverse_view/folder' cannot be created in a document with origin 'http://localhost:9876' and URL 'http://localhost:9876/context.html'.
	Error: Failed to execute 'pushState' on 'History': A history state object with URL 'http://localhost:8081/traverse_view/folder' cannot be created in a document with origin 'http://localhost:9876' and URL 'context.html'.
	    at Error (native)
	    at .<anonymous> (base/patterns/structure/js/views/app.js?ff70ad68d354c6623359d2d80a2550688c292c7d:152:26)
	    at Backbone.Paginator.requestPager.extend.pager (base/patterns/structure/js/collections/result.js?e1064bfadbd9ff184afcbad118a34744db759a76:60:12)
	    at Backbone.Model.extend.folderClicked (base/patterns/structure/js/navigation.js?e2f091ea37063d4cf66fae3ad79264256189db57:20:27)
	    at Backbone.View.extend.itemClicked (base/patterns/structure/js/views/tablerow.js?7110466d08b4135a2b29f473c58b07c7819feeeb:97:40)
	    at HTMLTableRowElement.$event.dispatch (base/lib/jquery.event.drag.js?5c512eae06a31d648141426af3ba2190a8a35572:374:19)
```

metatoaster:
> Tests pass - except tests in chrome and firefox on my local machine, but this might have some unknown system brokenness on my machine. See below, maybe someone has an idea?

It's part of the security checks a browser do on the `History.pushState()` API; as Karma is listening on `localhost:9876`, the test case uses `localhost:8081` which gets pushed, breaking things.

The other failure I found is due to my stub code for the pushState tests needs to also stub against the actual history object as it failed to be stubbed due to changes to various APIs over time (see discussion at sinonjs/sinon#719).

I got the fix done in 0c5591a7777b53989ac87146d5b450deaef36d80 (pushed into a separate branch `pr-651-ff-tests`). Feel free to bring that into this pull request if you wish, or I can do a pr after this one gets through.

thet:
@vangheem do you mind to look over this PR?
Controversial could be this: ``typeToViewAction``, tooltip pattern for all buttons  (requires a ``registry.scan`` for each actionmenu entry, if there are buttons instead of dropdowns), modal option which also need some changes in modal pattern).

thet:
I'm having troubles customizing individual files of the structure pattern in my own project. Customizing for example ``actionmenu.js`` ``z3c.jbot`` doesn't work apparently, because z3c.jbot seem not to support resources within a browser resource directory. Also, I cannot simply override a requirejs configuration entry, since all resources are accessed relative via ``mockup-patterns-structure-url``.
This makes me think, if we should add for each pattern resource file an individual config.js entry. This would make it much easier to override individual files (instead of extending jbot to support browser resource directories). The drawback would be, that we have a lot more entries in config.js and also need to update the plone registry for that.

vangheem:
As for the /view implementation, maybe a better would would just be to have the ajax vocabulary return an additional attribute, "viewUrl"? That way we don't have to deal with all these odd configuration solutions.

Otherwise, seems okay.

thet:
This PR is getting bigger and bigger...

The latest commit needs another review, because it again changes the UI in a way, which might be controversial:

https://github.com/plone/mockup/pull/651/commits/b80a592c715bd3d28ec5edc0f774ae884f5d547c
> Remove the checkbox and the actionmenu from the breadcrumbs bar for the current active folder to simplify the structure pattern.
> The actionmenu contained redundant actions (cut, copy, paste) and selecting the current folder is possible one level up.

IMO this simplifies the UI by reducing the amount of buttons. we could provide an option to enable/disable it, but I'm for leaving it out because I don't see much value in it.

I remember @ebrehault saying, he is teaching his clients especially this button?
@vangheem what do you think?

vangheem:
I'm fine with it; although, I'm pretty sure I added it because someone complained about having to go up to the parent :)

I think the real "fix" is perhaps to have a ".." entry for the first item in the results that points at the parent; however, it is more difficult to implement.

jensens:
What about merging this and then go further from here? Otherwise it feels like this one becomes too big.

khanstark:
If its fine, should I merge this?

thet:
@khanstark better not - I'd prefer to have that merged by a Plone core developer, who knows about these changes.

jensens:
i think this is a major step forward and so i merge now. there is still room for improvements, ok, but smaller steps are better :)

PR #652, opened 2016-03-31
Moment Updates

thet:
~~- Update the moment library to ``2.12.0``.~~
- Fix fakeserver ``relateditems-test.json`` response to return ISO dates for ``CreationDate``, ``ModificationDate`` and ``EffectiveDate``, as they really do in Plone.
  This resolves a moment deprecation warning in structure examples.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,10 @@ Incompatibilities:
 
 New:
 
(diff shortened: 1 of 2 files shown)


vangheem:
1 test failure. Merge when fixed!

thet:
Moment.JS doesn't support relative dates (YYYY MM DD) anymore. Only ISO dates. If we want to keep relative support, we probably need to keep at the 2.10.x versions.

vangheem:
I don't think we need to support non-iso dates so we're fine. Thank you.

PR #656, opened 2016-04-06
fix paste button not working

vangheem:


thet:
That was actually on my TODO list!
Tnx for fixing.

PR #662, opened 2016-04-12
Fixing problems with wrong relative date translations for norwegian sites

martior:
moment.js has a different language code for Norway then Plone uses. Pat uses nb, but Plone uses no for most Norwegian translations. This causes a problem with the relative dates showing as Asian characters. This is a possible solution, but my knowledge of mockup development is limited, so I have not been able to test this. 

This pull request is mostly for discussion and not meant to be merged as is.

### DIFF
--- a/mockup/patterns/moment/pattern.js
+++ b/mockup/patterns/moment/pattern.js
@@ -109,7 +109,12 @@ define([
       if (!date || date === 'None') {
         return;
       }
-      moment.locale([(new i18n()).currentLanguage, 'en']);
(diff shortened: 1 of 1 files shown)


jensens:
I just found this one-year-old PR. I think this one is still valid.
@martior hence you branched away in your personal space I can't rebase and add a changelog entry. If you can do so and Travis is passing we can finally merge it.

thet:
Replaces by: https://github.com/plone/mockup/pull/828

PR #668, opened 2016-05-05
Fix issue where using filter when paging would not work

vangheem:


davilima6:
Fixes plone/Products.CMFPlone#1562

PR #670, opened 2016-05-16
Update package.json dependencies

thet:
Except less which has incompatible changes since 2.0 (less.parse).


jensens:
looks like travis does not like it.

PR #676, opened 2016-05-26
WIP Related Items Optimizations

thet:
# Related Items

  - [x] Make "search" and "browse" mode user-selectable via buttons.
  - [x] Remove obsolete tree widget, use "browse" mode instead.
  - [x] Let "search" mode search in currentPath recursively. The current path can be selected in "browse" mode.
  - [x] Show only selectable items in "search" mode, if defined.
  - [ ] Show only selectable and folderish items in browse mode, if selctable items are defined.
  - [ ] Allow optional image and file upload (especially useful for image and file reference widgets).
  - [x] Allow configuration of favorites based on registry path entries and show them as dropdowns to quickly navigate to these locations.

# Structure Pattern
- [x] Add ``patternClasses`` option for action menu items to be able to set abitrary patterns (or classes) on the anchor tag of the menu item.
- [x] Prevent popovers to be closed when clicking on non-visible elements which still use screen space (those with ``visibility: hidden`` or ``opacity: 0``).
  That in addition to elements, which are not visible at all and do not use screen space (``display: none``).
  Prevents upload form in structure pattern to be closed when opening the file selection tool.

# Popover
- [x] Add option ``allowPathSelection`` to the upload widget to explicitly disable the pat-relateditems path selection.

# Contentloader
- [x] Contentloader pattern: Remotely loaded HTML content is wrapped in a ``div`` element, to allow jQuery to find also the first element. jQuery starts to search at it's first child element.

The scope of this PR is getting bigger. Sorry!
Reason is, this cannot be merged yet, as this breaks a lot of core tests.
I'm short on time for fixing bugs and adding features for an ongoing client project, so I keep developing this stuff in here.
I'm doing it as generic as possible, so these fixes and features are useful for all, IMO.


thet:
Conflicts with: 8e3bac7d922c25fb0759a29031a113674846fcad

thet:
Rebased: https://github.com/plone/mockup/pull/696

PR #679, opened 2016-05-31
Plip 1483 retina image scales

didrix:
https://github.com/plone/Products.CMFPlone/issues/1483

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -118,6 +118,9 @@ New:
 - Add an option to set the path operator in QueryHelp and tinymce pattern.
   [Gagaro]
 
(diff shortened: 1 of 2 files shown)


vangheem:
For b/w compatibility, could you provide a default? Something like `self.$el.attr('data-modal-srcset') || ''`

didrix:
@vangheem thanks. I assumed 'undefined' would result in an empty string, but is doesn't :|

mauritsvanrees:
This is part of a PLIP, but is fine to consider on its own, as it does not introduce any new features, it just prepares for a new feature.

PR #684, opened 2016-06-05
Escape input into select2 widget

vangheem:


mauritsvanrees:
The tests fail. Some are not clear, just 'false is not true', but others indicate an unexpected change in url:

http://jenkins.plone.org/job/pull-request-5.1/312/testReport/junit/Products.CMFPlone.tests.testResourceRegistries/TestScriptsViewlet/test_request_resources/
```
-  'src': 'http://localhost:47240/plone/++resource++foo.js'}
?                 ^ ---    ------

+  'src': 'http://nohost/plone/++resource++foo.js'}
?                 ^
```

vangheem:
The error doesn't have anything to do with mockup. Not until the js/css is rebuilt in CMFPlone does any of the mockup changes really matter much in fact.

thet:
Wouldn't it be necessary to use the escape function also on line ``225`` ( https://github.com/plone/mockup/pull/684/files#diff-7df2098dccd749a22bb1f7cd0706dd8cR225 ), where:
```
data.push({id: this, text: this});
```
?
Otherwise LGTM and should be merged when appropriate action on my comment is done.

thet:
Eventually these also need to escaped:
https://github.com/plone/mockup/blob/cbfd3fdb41027ecc6264ae1368a8a26dcb743697/mockup/patterns/querystring/pattern.js#L544
https://github.com/plone/mockup/blob/cbfd3fdb41027ecc6264ae1368a8a26dcb743697/mockup/patterns/querystring/pattern.js#L549
https://github.com/plone/mockup/blob/cbfd3fdb41027ecc6264ae1368a8a26dcb743697/mockup/patterns/querystring/pattern.js#L563
https://github.com/plone/mockup/blob/cbfd3fdb41027ecc6264ae1368a8a26dcb743697/mockup/patterns/querystring/pattern.js#L595
https://github.com/plone/mockup/blob/cbfd3fdb41027ecc6264ae1368a8a26dcb743697/mockup/patterns/querystring/pattern.js#L606
https://github.com/plone/mockup/blob/cbfd3fdb41027ecc6264ae1368a8a26dcb743697/mockup/patterns/querystring/pattern.js#L887
https://github.com/plone/mockup/blob/cbfd3fdb41027ecc6264ae1368a8a26dcb743697/mockup/patterns/querystring/pattern.js#L903

https://github.com/plone/mockup/blob/cbfd3fdb41027ecc6264ae1368a8a26dcb743697/mockup/patterns/resourceregistry/js/overrides.js#L66
https://github.com/plone/mockup/blob/cbfd3fdb41027ecc6264ae1368a8a26dcb743697/mockup/patterns/resourceregistry/js/overrides.js#L108
https://github.com/plone/mockup/blob/cbfd3fdb41027ecc6264ae1368a8a26dcb743697/mockup/patterns/resourceregistry/js/overrides.js#L135
https://github.com/plone/mockup/blob/cbfd3fdb41027ecc6264ae1368a8a26dcb743697/mockup/patterns/resourceregistry/js/overrides.js#L143

This are the resouces where I found some equivalent and appropriate looking ``data.push`` together with the use of select2.

Would it be possible to always provide an empty data array from select2 and overwrite the push method, which then always does the escaping? Just an idea....

vangheem:
Yes, it would :) I'll update

vangheem:
I don't think the querywidget needs to be escaped. tTese are selections for select2, users can't input raw text into select2 widget on the querystring pattern.

vangheem:
The problem is only because we allow free text input in the select2 pattern(like new tags, new contributors, etc). This does not happen with querystring.

thet:
OK, same for resource registry.
I'll merge!

PR #685, opened 2016-06-06
update bower dependencies

thet:
Update safe bower.json dependencies except backbone which tests would fail.

Reason was to update tinymce....


thet:
Ref: https://github.com/artursmirnov/tinymce_builded/pull/9

PR #689, opened 2016-06-17
Fix issue where TinyMCE selection position was not updated on modal re-shown

datakurre:
@vangheem After https://github.com/plone/mockup/commit/efd02e50904f55299eb7f00eddc6c9fcf762ffb1 have you still been able to add multiple new links with TinyMCE without reloading the edit page?

For me, because "init" is called only when the link modal is shown for the first time, all link actions affect the element which was selected for the first link action (e.g. adding in-line images will all go into the same place regardless of the cursor position after the first addition). Moving your added code from init to initData seem to have fixed that for me (because initData is called every time when modal is shown), but I'm not yet sure if it has any side effects.

### DIFF
--- a/mockup/patterns/tinymce/js/links.js
+++ b/mockup/patterns/tinymce/js/links.js
@@ -409,38 +409,6 @@ define([
       self.dom = self.tiny.dom;
       self.linkType = self.options.initialLinkType;
       self.linkTypes = {};
-
-      self.data = {};
-      // get selection BEFORE..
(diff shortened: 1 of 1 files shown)


vangheem:
Ahh yes, I think I remember this.

I've been working on a fork of the tinymce pattern for a bit now for various reason and haven't had the time to try and merge some of it back. 

In my fork it seems I moved the code to the initElements method but this should work about the same.

PR #694, opened 2016-08-05
Fix to check undefined with typeof foo === 'undefined'

datakurre:
Not sure why, but in my webpack-built bundles checks like foo !== undefined in querystring may fail (pass for undefined), but typeof foo !== 'undefined' works correctly.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -55,6 +55,10 @@ New:
 
 Fixes:
 
+- Fix issues where querystring widget was broke due to issues with
+  checks for undefined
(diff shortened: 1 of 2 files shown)


datakurre:
Well, now when I look that pattern, it looks like undefined is defined as select2. I wonder, how that works for anyone, or maybe there are differences between browser (AFAIK it should not be possible to re-define undefined in modern browsers, but ... )

vangheem:
the `typeof` `undefined` syntax never makes any sense to me.

PR #702, opened 2016-08-26
Use fork of grunt-sed

gforcada:
So seems that grunt-sed is used in quite some places, so it can not be removed as my original idea was.

Fortunately there is a pull request to fix the dependency problems https://github.com/jharding/grunt-sed/pull/17 *and* npm allows to define versions as github URLs  (even so short that you don't need to say that they come from github).

Long story short: I forked the pull request's fork on collective (just to prevent that the fork removes the commits) so now jenkins and travis both run fine :-)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,6 +13,8 @@ Fixes:
 - Update outdated links in Learn.md  
   [staeff]
 
+- Use github fork of grunt-sed and remove unused task.
+  [gforcada]
 
 2.3.0 (2016-08-19)
 ------------------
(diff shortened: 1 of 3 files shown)


gforcada:
To make testing this package even more funny travis fails to run (while it has been working fine) and jenkins starts working (which was failing up to now).

thet:
please do this version fix and any necessary changes also in: https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/_scripts/compile_resources.py#L20

after that, and if tests go green for jenkins, this should be ready to merge.

PR #711, opened 2016-09-17
tooltip cleanup and fixes

thet:
Make ``pat-tooltip`` useable by it's own by including the necessary less files and reuse that one in other patterns.
Allow configuration of ``placement`` parameter.

Without this fix, pat-tooltip cannot be easily used - except the bootstrap tooltip less files are also included.
What I do not fully understand is the ``tooltips`` classes from the tooltip less file, which I removed. That markup seems not to be used, although it's also used in plonetheme.barceloneta: https://github.com/plone/plonetheme.barceloneta/blob/master/plonetheme/barceloneta/theme/less/tooltip.plone.less

However, what I would really prefer is to remove the pat-tooltip from here alltogether and use instead the one from patternslib: http://patternslib.com/tooltip/
That should be an easy replacement, because It seems that mockup-patterns-tooltip wasn't exposed to be used by integrators anyways...

Any comments are welcome!
@vangheem @pilz @plone/framework-team

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -88,6 +88,10 @@ New:
 
 Fixes:
 
+- Make ``pat-tooltip`` useable by it's own by including the necessary less files and reuse that one in other patterns.
+  Allow configuration of ``placement`` parameter.
+  [thet]
+
 - Update outdated links in Learn.md
   [staeff]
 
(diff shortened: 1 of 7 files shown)


thet:
Merging it, as this won't break anything. Switch to patternlib tooltip is not in this scope and can be done sometime later.

Issue #706, opened 2016-09-06
select2 widget broken behavior with amp&;


tiberiuichim:
A lot of problems when an editor inserts a tag with an ampersand:
- not possible to delete this tag with the ampersand
- the tags are split wrong
  etc

I've tracked the problem to this line in /mockup/patterns/select2/pattern.js

```
data.push({id: utils.escapeHTML(this), text: utils.escapeHTML(text)});
```

If the escapeHTML call is not used at all, things work just fine.

jensens:
i can confirm this.

petschki:
jep here too ... for example if you want to add a keyword "Q&A" it get's splitted by ";" in `["Q&amp", "A"]` values. Completely removing this would lead to XSS vulnerability. I'd suggest to introduce a more restrictive utils method:

```
  var removeHTML = function(val) {
    return val.replace(/<[^>]+>/ig, "");
  };
```
and implement it instead of `utils.escapeHTML` in `pat-select2`

ping @thet @vangheem any thoughts ?

thet:
Fixed with: https://github.com/plone/mockup/issues/706

PR #722, opened 2016-11-07
fix for #706

petschki:
introduce new utility method `removeHTML` and implement it to the
select2 pattern. add tests for html manipulation and select2 values.


jensens:
@thet may you review this? Is this the source of the problems in our aut project with the ``&`` dividing vocabs?

thet:
Looks good to me and solves a problem we also have.
I just added another test to check if html isn't escaped as it was before: "do not html-escape input"

PR #728, opened 2017-01-02
Add better support for TinyMCE importcss-plugin configuration through pattern

datakurre:
Currently configuring importcss-plugin is very limited, because even the plugin supports regexp filters here and there, those cannot be passed through pattern configuration. Therefore, I suspect, that almost nobody else has used importcss_groups and importcss_selector_filter -options.

This change would add implicit RegExp support for importcsss_selector_filter and filter part of each importcss_groups-value. I believe, this is backwards compatible, because replacing "myprefix" with RegExp("myprefix") should not have much effect.

http://archive.tinymce.com/wiki.php/Configuration:importcss_selector_filter
http://archive.tinymce.com/wiki.php/Configuration:importcss_groups

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -99,6 +99,11 @@ New:
 - Add select2 result class based on value of option so it can be styled
   [vangheem]
 
+- Add to interpret TinyMCE ``importcss_selector_filter`` and ``filter`` value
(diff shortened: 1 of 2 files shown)


vangheem:
lgtm

Issue #734, opened 2017-02-02
Why is the query default path to the navigation root and not the portal root?


Gagaro:
Hi,

I'm talking about https://github.com/plone/mockup/blob/master/mockup/js/utils.js#L102-L114 and https://github.com/plone/mockup/blob/master/mockup/patterns/relateditems/pattern.js#L192-L196.

I have a lot of navigation root in my site, and I want to be able to link to another part of the site, which isn't possible in the current state. It worked ni 5.0.

So, is there a reason why those are from the navigation root or can we change them to `absolutePath`?

jensens:
Closing as outdated. Mockup was completely rewritten for Plone 6 (jQuery/Backbone/RequireJS → ES6/Patternslib). Old relatedItems query default path.
If you think this issue is still relevant in the new implementation, please reopen it. Thank you for the original report!

PR #736, opened 2017-02-03
Query from portal path rather than from navigation root

Gagaro:
Related to #734.

The `rootPath` is actually already set by default on the navigation root, so the query can be made on the portal root without issue to allow for more customization if needed :
https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/patterns/settings.py#L118-L119

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -133,6 +133,9 @@ Fixes:
 - Fix conflict between upload and relateditem browse button.
   [Gagaro]
 
+- Query from portal path rather than from navigation root.
+  [Gagaro]
+
 
 2.3.0 (2016-08-19)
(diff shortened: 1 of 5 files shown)


jensens:
@thet may you have a look at this one?

Gagaro:
Do not merge. This breaks because of the base_path being removed from the query path in plone.app.content (and being the navigation root by default) : https://github.com/plone/plone.app.content/blob/master/plone/app/content/browser/vocabulary.py#L219-L220.

I'll see if it would be possible for these queries to be configurable instead of juste changed to absolute.

Gagaro:
Closing in favor of #742

PR #741, opened 2017-02-09
Pickadate optimizations - backport from master

thet:
- Remove the now obsolete footer buttons as a whole from the date picker.
- Add options today and clear to hide those buttons when set to false.
- Use display: inline-block instead of problematic float:left.


thet:
that needs also changes in plone.app.widgets. I close this one and focus for Plone 5.1

PR #742, opened 2017-02-10
feat: add an option for the path operator in QueryHelper and tinymce pattern

Gagaro:
Related to #734 and #736.

It allows to change the query to be done on the portal root for example.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -115,6 +115,9 @@ New:
 - Add ``defaultSortOn`` option in ``livesearch`` pattern.
   [Gagaro]
 
+- Add an option to set the path operator in QueryHelp and tinymce pattern.
+  [Gagaro]
+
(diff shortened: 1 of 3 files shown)


thet:
@Gagaro for the feature and @ebrehault thanks for the merge - but @Gagaro - your changelog reads:

```
- Add an option to set the path operator in QueryHelp and tinymce pattern.
```

Shouldn't it be:

```
- Add an option to set the path operator in QueryHelp and relateditems pattern. 
```

?

Gagaro:
Yes indeed, I actually needed this for the related widget inside tinymce, so my brain mixed this.

I fixed it in https://github.com/plone/mockup/commit/1bcad9ff37ad24474b498d8bc5d47c097c3fa86e, thanks for the review :+1: .

PR #744, opened 2017-02-24
fix datepicker markup

jensens:
see plone/Products.CMFPlone#1953


jensens:
now it looks like so:
screenshot from 2017-02-27 10 26 37

thet:
looks good to me!

PR #748, opened 2017-03-08
Structure: if emtpy title then show ID instead

jensens:


frapell:
@jensens This makes a lot of sense, however should we also include some additional way of being very clear that this is not the title of the element? I imagine editors not realizing that what they are seeing is not actually the element title?

Maybe a special class for the <em> or additional wording at the beginning? (something like <em><span>id: </span><%- id %></em>)

jensens:
@frapell I so did the ``<em><%- id %></em>`` - if this is not enough I am open for better ideas!

frapell:
@jensens Ah! Sorry, I thought the title was also wrapped in ```<em></em>``` Is this ready to be merged in?

jensens:
@frapell yes, all ready.

PR #751, opened 2017-03-18
Do not use relative paths for imports in less files.

thet:
Use the less variables for paths instead.
Fixes a case, where less files couldn't be found in Plone development mode.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,11 @@ New features:
 
 Bug fixes:
 
+- Do not use relative paths for imports in less files.
+  Use the less variables for paths instead.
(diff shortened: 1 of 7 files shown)


CHANGES.rst
```diff
@@ -14,6 +14,15 @@ New features:
 
 Bug fixes:
 
+http://localhost:8080/manage
+
+http://localhost:8080/manage
+
```
ale-rt: @thet is this double "http://localhost:8080/manage" something you left by mystake?

thet: absolutly.. I'll fix it.

PR #752, opened 2017-03-28
more related items widget optimizations and refactorings

thet:
- Add new mode "auto", which automatically sets ``search`` mode when a searchterm is present, otherwise ``browse`` mode.
- Show a "One level up" button in the result set in browse mode.
- Show folder icons in in the result set front of items instead a arrow behind the items.
- Filter out non-selectable and non-folderish items in the result set when in browse mode.
- Add option to scan the selected list of items for other patterns.
- Add option for contextPath - objects with this path will not be selectable. This prevents the object where the relation is set on to from being selected and self-referenced.


thet:
Until now these PR doesn't break anything....
But I propose even some more radical changes:

- Remove the "browse" and "search" buttons completly. I think they are not necessary anymore. Only if there is a need to allow users to only search within a specific path and not break out of it...

- Remove the xyzTemplateSelector options. Is anyone using templates from the DOM?

- Instead of passing a template string, allow to pass a resource path to a template.

jensens:
>  Remove the "browse" and "search" buttons completly. I think they are not necessary anymore. Only if there is a need to allow users to only search within a specific path and not break out of it...

Well, I would keep them for now.

> Remove the xyzTemplateSelector options. Is anyone using templates from the DOM?

+1 I never saw them used out in the wild.

> Instead of passing a template string, allow to pass a resource path to a template.

+1

PR #755, opened 2017-04-06
More structure widget optimizations

thet:
- Show ineffective label in folder contents for not yet effective and published content, likewise it's done with expires.
  Show effective and ineffective label styled as bootstrap badges.

- Show "Description" below title, if it's set in ``availableColumns`` and ``activeColumns`` to save some screen space.

- Add default plone color less variables for a more consistent design.
  They will be overwritten by values set by Plone or integration projects.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,10 @@ Breaking changes:
 
 New features:
 
+- Add default plone color less variables for a more consistent design.
+  They will be overwritten by values set by Plone or integration projects.
+  [thet]
+
 - Structure widget:
   - Show ineffective label in folder contents for not yet effective and published content, likewise it's done with expires.
     Show effective and ineffective label styled as bootstrap badges.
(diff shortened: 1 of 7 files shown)


davisagli:
"Ineffective" is an odd word to use for this in English. It means "this is not successful at producing its intended effect."

"Before publishing date" might be the clearest given that the related field on the edit form is called Publishing Date. But it's a bit long, and perpetuates the confusion between effective date and workflow state. I'm not sure what to recommend. "Not yet effective" would probably be unclear, but would be better than "Ineffective"

thet:
@davisagli thanks for checking the wording. I was also not sure about this.

I agree, "Not yet effective" is better than "ineffective". But all "effective date" related UI elements in are translated to something with "Published" in English ("Publishing Date" or "Publication Date").

I think your suggestion of using "Before publishing date" is the best and we should use it.

thet:
Looks like this:

screen shot 2017-04-13 at 13 21 41

PR #756, opened 2017-04-11
Fixed path to tooltip less files.

mauritsvanrees:
This gave an ugly site in develoment mode when editing the loggedin bundle css.
Fixes https://github.com/plone/Products.CMFPlone/issues/1843.


thet:
awesome! works for me.

PR #759, opened 2017-04-26
Fix issue where formunloadalert pattern raised initialization error for modals

datakurre:
I was annoyed by formunloadalert pattern raising errors into JS console when modals were opened.

The broken code in question was years old. I have no idea, what it is supposed to do and if it still really works: https://github.com/plone/mockup/commit/22ead2779a0897442db2977076714b92fff9ea16

Yet, I found that what was previously data('pattern-modal') is nowadays data('patternPloneModal'). I assume this change is related to making mockup Patternslib compatible. @thet might know

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -88,6 +88,9 @@ Bug fixes:
 - fix localization of "Open folder" link title in related items pattern
   [datakurre]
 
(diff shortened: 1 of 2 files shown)


thet:
@datakurre - I don't now, but LGTM.

PR #761, opened 2017-04-27
New pattern mockup-patterns-structureupdater

thet:
- New pattern ``mockup-patterns-structureupdater`` to update title and description depending on the current context on Plone's folder contents page.
- Structure widget: Trigger ``context-info-loaded`` on body to be able to listen to the event outside the pattern.

### DIFF
--- a/mockup/patterns/structure/js/views/addmenu.js
+++ b/mockup/patterns/structure/js/views/addmenu.js
@@ -14,7 +14,7 @@ define([
     initialize: function(options) {
       var self = this;
       ButtonGroup.prototype.initialize.apply(self, [options]);
-      self.app.on('context-info-loaded', function(data) {
+      $('body').on('context-info-loaded', function(data) {
         self.$items.empty();
         _.each(data.addButtons, function(item) {
           var view = new ButtonView({
--- a/mockup/patterns/structure/pattern-structureupdater.js
+++ b/mockup/patterns/structure/pattern-structureupdater.js
@@ -1,0 +1,40 @@
+/* Structure Updater pattern.
+ *
+ * Options:
+ *    titleSelector(string): CSS selector to match title elements for the current context ('').
+ *    descriptionSelector(string): CSS selector to match description elements of the current context to be updated ('').
+ *
+ * Documentation:
+ *    The Structure Updater pattern updates the information of the current context after a user navigated to a new folder.
+ *
+ */
+
+define([
+  'jquery',
+  'pat-base'
+], function($, Base) {
+  'use strict';
+
+  var StructureUpdater = Base.extend({
+    name: 'structureupdater',
+    trigger: '.template-folder_contents',
+    parser: 'mockup',
+    defaults: {
+      titleSelector: '',
+      descriptionSelector: ''
+    },
+
+    init: function() {
+
+      $('body').on('context-info-loaded', function (e, data) {
+        $(this.options.titleSelector, this.$el).html(data.object && data.object.Title || '&nbsp;');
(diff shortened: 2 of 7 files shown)


mockup/patterns/structure/js/views/addmenu.js
```diff
@@ -14,7 +14,7 @@ define([
     initialize: function(options) {
       var self = this;
       ButtonGroup.prototype.initialize.apply(self, [options]);
-      self.app.on('context-info-loaded', function(data) {
+      $('body').on('context-info-loaded', function(data) {
```
thet: the event is now triggered on the body, so that I can listen outside of the pattern for it.
(the event didn't bubble up, don't know why..)

mockup/patterns/structure/pattern-structureupdater.js
```diff
@@ -0,0 +1,40 @@
...
+
+define([
+  'jquery',
+  'pat-base'
+], function($, Base) {
+  'use strict';
+
+  var StructureUpdater = Base.extend({
+    name: 'structureupdater',
+    trigger: '.template-folder_contents',
```
thet: please note the trigger. it's what's in the body class, set by https://github.com/plone/plone.app.layout/blob/master/plone/app/layout/globals/layout.py#L187
That way, I do not need to update this method in plone.app.layout to add a ``pat-structureupdater`` class or so.
I think it's fine, even if it doesn't follow closely the naming conventions.

mockup/patterns/structure/pattern-structureupdater.js
```diff
@@ -0,0 +1,40 @@
...
+    parser: 'mockup',
+    defaults: {
+      titleSelector: '',
+      descriptionSelector: ''
+    },
+
+    init: function() {
+
+      $('body').on('context-info-loaded', function (e, data) {
+        $(this.options.titleSelector, this.$el).html(data.object && data.object.Title || '&nbsp;');
```
thet: title and description are cleared, if the portal root is reached. there, no context info ajax call is done, so I do not get the portal name there.
I'm ok with that.
the ``&nbsp;`` is for keeping the line height of the title/desc intact, so folder contents doesn't "jump around" when navigating to the portal root.

mockup/patterns/structure/pattern-structureupdater.js
```diff
@@ -0,0 +1,40 @@
...
+        $(this.options.descriptionSelector, this.$el).html(data.object && data.object.Description || '&nbsp;');
```
thet: actually, the description isn't shown on the folder contents page. I could remove that.

datakurre:
@thet was there reason not to implement this directly in structure pattern?

thet:
@datakurre - I wanted to have it decoupled from the structure pattern. The title and description are outside the scope of the structure pattern and those two are only two aspects of UI elements, which should potentially be updated (whereas description is not available in standard Plone folder_contents).
In concrete projects, there might be other ones too - or someone also needs to have the section viewlet and breadcrumbs viewlet be updated. If that's the case, then you can more or less easily overwrite the RequireJS configuration of that pattern and use a own implementation.

Regarding the section and breadcrumbs viewlet - I didn't want to include their update in the pattern because:
- A costly AJAX call would be needed.
- A view-wrapper would be needed too - which would introduce a new concept (wrapping viewlets in views), where we already have a better one - tiles. But they are not yet approved for Core...

Makes sense? I do not have a very strong opinion on how I implemented that pattern...

datakurre:
@thet This makes sense now. Thanks. I agree with what you said. Eg. If I have time to integrate this for our sites, I would probably write a custom pattern, which listens the event and replaces content with layout tiles for selected context/

PR #773, opened 2017-06-12
Theme filemanager searching feature

b4oshany:
### Features Implemented

-  Search for files and opening the file upon click
-  Search for text within files and opening the file upon click
image
image

### UI Elements Added

- Bootstrap dropdown element
- Link (a tag) element


mockup/patterns/relateditems/pattern.js
```diff
@@ -98,11 +98,11 @@ define([
...
-  'text!mockup-patterns-relateditems-url/templates/breadcrumb.xml',
-  'text!mockup-patterns-relateditems-url/templates/favorite.xml',
-  'text!mockup-patterns-relateditems-url/templates/result.xml',
-  'text!mockup-patterns-relateditems-url/templates/selection.xml',
-  'text!mockup-patterns-relateditems-url/templates/toolbar.xml',
+//  'text!mockup-patterns-relateditems-url/templates/breadcrumb.xml',
+//  'text!mockup-patterns-relateditems-url/templates/favorite.xml',
+//  'text!mockup-patterns-relateditems-url/templates/result.xml',
+//  'text!mockup-patterns-relateditems-url/templates/selection.xml',
+//  'text!mockup-patterns-relateditems-url/templates/toolbar.xml',
```
datakurre: @b4oshany These have probably been forgotten here and are cause for the test errors. But why did you need to disable these imports in your setup in the first place?

mockup/js/ui/views/dropdown.js
```diff
@@ -0,0 +1,58 @@
...
+      ButtonGroup.prototype.initialize.apply(this, [options]);
+
+      this.on('render', function() {
+        this.renderTitle();
+      }, this);
+    },
+
+    renderTitle: function() {
+      var title = this.options.title;
+      if(this.options.title === undefined){
```
datakurre: @b4oshany This might be nitpicking, but maybe

```if(this.options.title === undefined){```

should be consistently written with spaces like

```if (this.options.title === undefined) {```

mockup/patterns/relateditems/pattern.js
b4oshany: It was breaking require js when building the compiled file on the resource registry page.

datakurre: That's strange. Possibly it is related to other TTW build failures when "++plone++unique" was introduced in late 5.0. That broke relative URLs in build.

datakurre:
Thanks. Looks good. Once I've tried this out myself once more, I'll merge this.

datakurre:
@b4oshany One minor issue. The toolbar looks a bit broken now:

* find-menu is placed like it is a part of a button group, but it's distinct of it because of rounded corners
* rename button has sharp left edge

I know these are minor, and you were about to return to UI related issues later, but these make the editor look a bit broken now. Once these are fixed, this feature would already be good to be merged and released (IMO).

datakurre:
Now that drag and drop is also in this pull, this depends on https://github.com/plone/plone.resourceeditor/pull/19

datakurre:
@b4oshany This looks good now. This can be merged as soon as we get dependencies for this merged. A few notes still to consider later:

* Drag and drop may still fail silently and move look completed when file is moved in inspection mode of a read only filesystem theme. That is rare use case so not priority.

* I feel it unintuitive that search dialog does NOT close when clicking on a result. But if it would close, it would also require saving its state so that the old results could be seen by just clicking find. Maybe. So, not trivial in UX wise.

datakurre:
This worked for me, new jqtree has been merged into CMFPlone and all checks pass. I'll merge.

PR #775, opened 2017-06-13
Queue inline validation calls

datakurre:
I've currently tested this only with dexterity objects. Probably can test with archetypes. Has Plone 5 anywhere formlib forms with this inline validation?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,12 +1,15 @@
 Changelog
 =========
 
-2.5.2 (Unreleased)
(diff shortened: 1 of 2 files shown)


jensens:
``zope.formlib`` is not officially supported by Plone 5 anymore. Just some addons are still using it.

datakurre:
It seems inline validation for Archetypes is not enabled by default on Plone 5, so I don't know, how to manually test this for archetypes.

thet:
LGTM

PR #782, opened 2017-06-26
Theme filemanager searching feature fixes

datakurre:
@b4oshany Small issue. If the "edited" theme is readonly, find menu actions get disabled. (Code in thememapper patterns calls disable for all button groups.) Iterating through all buttons to only enable find-menu looked complex in themepattern, so I propose simply disabling "disable" from find menu component.

### DIFF
--- a/mockup/patterns/filemanager/pattern.js
+++ b/mockup/patterns/filemanager/pattern.js
@@ -189,7 +189,8 @@ define([
           findinFilesView.triggerView
         ],
         id: 'find',
-        app: self
+        app: self,
+        disable: function() {}
       })
 
       self.views = [
--- a/mockup/patterns/thememapper/pattern.thememapper.less
+++ b/mockup/patterns/thememapper/pattern.thememapper.less
@@ -1,6 +1,6 @@
 @import "@{mockuplessPath}/ui.less";
 /* just be lazy here and include all of bootstrap... */
-@import "@{mockupPath}/filemanager/pattern.filemanager.less";
+@import "@{mockup-patterns-filemanager}";
 
 body.plone-toolbar-left-default {
 	padding-left: 0px;


mockup/patterns/filemanager/pattern.js
```diff
@@ -189,7 +189,8 @@ define([
           findinFilesView.triggerView
         ],
         id: 'find',
-        app: self
+        app: self,
+        disable: function() {}
```
datakurre: Disables disabling, because disabling "find" does not make sense in this context (other buttons may be disabled when the managed files are read only, but that does not prevent search).

mockup/patterns/thememapper/pattern.thememapper.less
```diff
@@ -1,6 +1,6 @@
 @import "@{mockuplessPath}/ui.less";
 /* just be lazy here and include all of bootstrap... */
-@import "@{mockupPath}/filemanager/pattern.filemanager.less";
+@import "@{mockup-patterns-filemanager}";
```
datakurre: It's best practice in mockup to use the aliases when they exist to make it easier to override resources in build tools (and I needed this to override Plone default version with this in my webpack dev env).

b4oshany: I think this was brought I via `git rebase`. I'll update this.

mockup/patterns/filemanager/pattern.js
b4oshany: Agreed

mockup/patterns/thememapper/pattern.thememapper.less
b4oshany: I just realised that this is a separate branch, I context switch and scanned through this. Yes, I agree.

PR #785, opened 2017-06-29
Add docs on how to customize the structure updater pattern

thet:
Here is some documentation on how to customize and overload the structure updater pattern.
It also applies on how to customize any pattern in general.

I've put it into mockup/mockup/patterns/structure - which doesn't follow our conventions but is probably a good place to put it to.
It's also written as .rst instead of MarkDown like the rest of mockups docs are written.
I used .rst because I know it better and I need some feedback anyways.

I'm not sure if this docs should live here or go to docs.plone.org - but I couldn't find an appropriate place there and I do not want to start a new chapter just to be able to put this documentation there....

any opinions or comments?

### DIFF
--- a/mockup/patterns/structure/README.rst
+++ b/mockup/patterns/structure/README.rst
@@ -1,0 +1,108 @@
+How to customize the Structure Updater pattern
+==============================================
+
+The ``mockup-pattern-structureupdater`` updates the title and description on the page - if available - when a folder is changed in the structure pattern. It's triggered by the class ``template-folder_contents``, which is available on the body tag when the folder contents page is opened. The pattern listens on the ``context-info-loaded`` event, which is triggered by the structure pattern.
+
+Often you will have customized layouts, where you need to update also other parts of the site when the user changes to another folder.
+
+
+Pattern configuration
+---------------------
+
+For simple cases, you can customize the CSS selector for the title and description via the options ``titleSelector`` and ``descriptionSelector`` for the pattern.
+This can be done by adding options for that pattern via the resource registry control panel under the "Pattern options" tab or via the ``registry.xml`` profile like so::
+
+  <record name="plone.patternoptions">
+    <value purge="False">
+      <element key="structureupdater">{"titleSelector": "h1.documentFirstHeading", "descriptionSelector": "footer"}</element>
+    </value>
+  </record>
+
+.. note::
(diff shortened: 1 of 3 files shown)


thet:
@agitator please re-review.

mockup/patterns/structure/README.rst
```diff
@@ -0,0 +1,108 @@
...
+    For the ``titleSelector`` and ``descriptionSelector`` you have to provide valid JQuery selectors.
+    Like with any CSS selector you can also specify mutliple selectors by seperating them via a comma sign.
+
+
+Pattern overloading
+-------------------
+
+If you need some more control, you can overload the pattern and provide your own.
+
+The pattern is registered in the RequireJS configuratio in ``mockup/mockup/js/config.js`` under the name ``mockup-patterns-structureupdater`` and under the path ``patterns/structure/pattern-structureupdater``.
```
gforcada: typo: configuration (missing an "n")

gforcada: the config.js path is mockup/js/config.js, right? mockup is there twice. One can assume that you mean inside the repository.

gforcada: Typos:
- ``configuration`` (misses the last n)
- config.js path should have only one mockup in it, right? the first mockup would be the repository, but if one looks at this, already inside mockup, would think that there is two nested folders named mockup in mockup's repository (or at least that's what happened to me)

thet:
@gforcada done! plz re-review.
What's your opinion on place and format of this docs?

gforcada:
@thet as for the text, it looks fine for me.

Regarding where it is placed, I don't see any other rst but the ones at the top level, all docs have been so far kept bundled inside the JS themselves. Is there a good reason to keep this one outside?

thet:
The reason I placed it within the structure pattern folder was, that wouldn't be a bad place if we hadn't already some kind of different structure and most importantly I wanted to get it done instead of thinking too long about the problem where to place it.

I should probably transfer it to MarkDown and place it alongside DEVELOP.md and LEARN.md.
I think it doesn't fit well into the autogenerated Mockup patterns overview page, as the text is too long.

gforcada:
@thet I wouldn't block the merge due to where to place the docs, but maybe it would be a good idea to create an issue for that and discuss it there. Probably with the @plone/documentation-team involved as well 👍

PR #790, opened 2017-07-27
Fix filtering of non-selectable items

thet:
Related Items: Fix filtering of non-selectable and non-browsable items, so that no empty list elements are contained.
Filtering behavior is: When browsing, show only folderish or non-selected, selectable items (but non-selectable, folderish items are greyed out).
When searching, show only selectable items, which were not already selected.
This fixes an issue where it was impossible to select items when many items were filtered out.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,7 +15,11 @@ Incompatibile changes:
 
 Bug fixes:
 
-- * Add item here *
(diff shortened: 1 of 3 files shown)


thet:
... so I found a better way ...
finishing my work on it...

thet:
done, and force-pushed

PR #794, opened 2017-08-08
Fix callback for sortable pattern

tomgross:
This PR fixes the drop-callback of the sortable pattern

https://community.plone.org/t/mockup-pattern-sortable-associating-a-callback-function-with-drag-and-drop-not-working-as-expected/3608

Since the pattern arguments always seem to be converted to strings, this patch looks up the method name in the global window space.

https://stackoverflow.com/questions/359788/how-to-execute-a-javascript-function-when-i-have-its-name-as-a-string

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,6 +15,9 @@ Incompatibile changes:
 
 Bug fixes:
 
+- Fix callback of sortable pattern
+  [tomgross]
+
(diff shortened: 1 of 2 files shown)


tomgross:
@thet Do you mind to review please?

thet:
LGTM!
I have updated the docs and comments to reflect your change.

PR #797, opened 2017-08-16
Related Items: Simplify templates a bit.

thet:


thet:
Looking forward for plone.restapi beeing in core and getting the relateditems widget completly rewritten.

PR #784, opened 2017-06-28
Contextual menu

b4oshany:
Added the following contextual menu to file tree for the theme mapper:

* New File
* New Folder
* Rename
* Delete

resized-contextualmenu

PR #803, opened 2017-08-25
Rapdio Store support

b4oshany:
* [x] Requires [pull request 784](https://github.com/plone/mockup/pull/784) to be merged first

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,9 @@ New features:
 
 - Added docs for the structureupdater pattern.
   [thet]
+- Added Rapido Store to thememapper pattern.
+  - Allow users to install rapido apps from other themes to the cotexted theme.
+  [b4oshany]
 
 - Related Items: Simplify templates a bit.
   [thet]
--- a/mockup/patterns/filemanager/js/addnew.js
+++ b/mockup/patterns/filemanager/js/addnew.js
@@ -19,6 +19,16 @@ define([
     events: {
       'click button': 'addButtonClicked'
     },
+    show: function () {
+      // Override the PopoverView.show method
+      var self = this;
+      // Call the parent method, PopoverView.show
+      PopoverView.prototype.show.call(self);
+      // Foxus the keyboard to the input field.
+      var $input = self.$('input');
+      $input.val("");
+      $input.focus();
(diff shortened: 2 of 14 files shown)


ebrehault:
@b4oshany @datakurre @thet I do not think it should be merged into the core, this should be done via an add-on.

`rapido.plone` could be that add-on: it would provide its own version of the thememapper resource and it would override the default Plone one.

b4oshany:
@ebrehault, I understand that not all Plone sites will have rapido installed and this will invalidate having rapido store inside of `plone.mockup`.  In cases like that, the rapido store button will be disabled or hidden from the Plone Theme Editor view. 

NB the rapido store backend is found in the `rapido.plone` add-on and the button on the Theme Editor page make RESTFUL requests to the store API.

### Questions about moving the mockup files for rapido store to `rapido.plone`:
* Isn't overriding the default thememapper resource a bit risky considering that there might be another package overriding it? 
* Isn't conflict resolution going to be a problem where changes in the default thememapper resource have to be updated in the overridden resource.
* Is it ok to break the intended culture to place interrelated patterns outside of plone.mockup?

ebrehault:
Well, I agree the resource registry does not manage very well overiding at the moment :)

So yes it is risky, yes it can produce conflicts, but still, we really cannot put anything specific to a given add-on into Plone core.

And regarding your last point, I think it is totally ok to have an external package providing mockup-based resources (mockup is just a regular dependency here, just like any other part of Plone core).

datakurre:
If Rapido is willing to maintain its own thememapper, it's ok for me. Although, there is a risk, that it will become a fork with further improvements, not backported into this upstream version.

Any ideas, how thememapper could be made dynamically extensible?

I recall that the thememapper page does include pattern configuration from registry, so in theory, it could read "plugins" from pattern options (generated from registry) and do dynamic require calls for them.

datakurre:
About mockup, I recall hearing that there was consensus in Plone Open Garden that mockup could be eventually split into reasonable size NPM packages to follow JavaScript ecosystem best practices.

jensens:
There was no activity on this for a while. I close this request for now. If you plan to work further on this please reopen and go on.

PR #810, opened 2017-09-22
Concatenate value for select multiple input in generic popover in the structure pattern

Gagaro:
The `remove tags` input in the tags popover of the structure pattern uses a `select multiple`.

`serializeArray()` produce one object for each value selected in this input. So only the last one is saved in the data (and posted to the view).

This PR concatenate the value to data if a value with the same name is already present (with a `,`, same as how the `pat-select2` with multiple works).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,9 @@ Bug fixes:
   Fixes #808.
   [thet]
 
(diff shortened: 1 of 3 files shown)


Gagaro:
Test added

PR #813, opened 2017-10-18
Fixed resource registry controlpanel overrides tab save button

obct537:
The override pattern was removing the "disabled" property, but the button was being disabled by the disabled CSS class, making the buttons unusable.


thet:
LGTM

PR #815, opened 2017-10-23
Added options to change sorting in relateditems

Gagaro:


Gagaro:
The tests failings seems to be unrelated and I have an error when I try to rebuild them, is there another way to re-do them?

jensens:
in order to make this effective, you also need to rebuild the bundle in Products.CMFPlone

PR #817, opened 2017-11-04
typo fix in error message

tkimnguyen:


tkimnguyen:
Do I need to make a similar change in plone.app.locales/widgets.pot to the matching msgid?

thet:
@tkimnguyen yes that would be needed.
That's the downside of using the actual message strings as message id's :(

tkimnguyen:
ok I made this PR to match: https://github.com/collective/plone.app.locales/pull/177

tkimnguyen:
(It's kind of weird for a core package to depend on a collective one...)

thet:
tnx!

PR #819, opened 2017-11-06
Allow more flexible anchors in TinyMCE

tomgross:
With the current implementation 2 kinds of elements are parsed for the anchor selction in TinyMCE

1. `a`  and `img` with either the `mceItemAnchor`  or the `mce-item-anchor` class present
2. elements set by the `anchorSelector` parameter (`h1,h2,h3` by default)

This PR changes the behavior of 1. It adds **ALL** elements with one of the classes  `mceItemAnchor`  or `mce-item-anchor` set to the list of available anchors.

The risk of breaking something should be very low. The advantage is that you can allow anchors for custom elements (with the `name` or `id` attribute) by setting the marker class.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,11 @@ New features:
   - Added options to change sorting.
   [Gagaro]
 
+- TinyMCE pattern:
+
+  - Make anchor handling more flexible
+  [tomgross]
+
 Bug fixes:
 
 - Structure pattern:
--- a/mockup/patterns/tinymce/js/links.js
+++ b/mockup/patterns/tinymce/js/links.js
@@ -224,7 +224,7 @@ define([
       self.anchorData = [];
       var node, i, j, name, title;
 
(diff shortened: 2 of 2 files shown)


tomgross:
@thet Can you please review?

frapell:
@tomgross Change looks super easy...

I do wonder why it was specifically set on ```a.mceItemAnchor,img.mceItemAnchor,a.mce-item-anchor,img.mce-item-anchor``` before... do you know if there is any limitation and it had to be that way?

If everything continues to work as before, with the addition of being able to specify more anchors, I'd say +1

tomgross:
@frapell Thanks for reviewing! These selectors are here since this functionality was introduced in https://github.com/plone/mockup/commit/e0c6430c337b1b4f909f8b47595b8bb37d43c786#diff-97b5e50dba79f4265ee43d67dddae373R299 Maybe @vangheem knows about it? I expect everything working as before with the change. I myself want to utilize the `mce-item-anchor` class to add custom anchors. I guess someone using this class unintentionaly is very unlikely.

frapell:
Yeah, I think you are right, and this selector should continue to match both a and img that have those classes, so should continue to work fine

PR #826, opened 2018-01-08
Related Items: Add dropdown with recently added items.

thet:
recently used items are stored in ``localStorage``. 

note: the related PR https://github.com/plone/plone.app.relationfield/pull/21 was superseded by 
https://github.com/plone/plone.app.widgets/pull/177

merge together:
https://github.com/plone/mockup/pull/826
https://github.com/plone/Products.CMFPlone/pull/2257
https://github.com/plone/plone.app.widgets/pull/177
https://github.com/plone/plone.app.relationfield/pull/22
https://github.com/plone/plone.app.z3cform/pull/84
https://github.com/plone/Products.Archetypes/pull/102

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,7 +10,8 @@ Breaking changes:
 
 New features:
 
-* add items here
+- Related Items: Add dropdown with recently added items.
+  [thet]
 
 Bug fixes:
 
@@ -42,7 +43,7 @@ New features:
 
   - Make anchor handling more flexible
   [tomgross]
-  
+
   - Mark special links
   - Do not mark anchors as special links
   [frapell]
--- a/mockup/js/utils.js
+++ b/mockup/js/utils.js
@@ -335,7 +335,23 @@ define([
   };
 
   var removeHTML = function(val) {
-    return val.replace(/<[^>]+>/ig, "");
+    return val.replace(/<[^>]+>/ig, '');
+  };
+
+  var storage = {
+    // Simple local storage wrapper, which doesn't break down if it's not available.
+    get: function (name) {
+        if (window.localStorage) {
+          var val = window.localStorage[name];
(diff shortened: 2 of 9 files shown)


thet:
@jensens - tnx for reviewing. I like the idea of adding the field name to the key - will do so. Although here all items are filtered to what is possible to add: https://github.com/plone/mockup/pull/826/files#diff-b4d083c1c196b28befec6eefebba38e1R338 - so your concern shouldn't be that much of a problem.

Regarding adding the user id - that came to my mind too, but I'm not sure if that is a problem too. Generally on one and the same computer cookies, form data, etc are normally not deleted by the user. Also other programs also provide a recently used items list. All those data is bound to the user which is logged in at OS level.
However, adding the userid to the key, likewise as proposed for the fieldname shouldn't be a problem.
What do you think?

jensens:
>  I like the idea of adding the field name to the key - will do so. Although here all items are filtered to what is possible to add: https://github.com/plone/mockup/pull/826/files#diff-b4d083c1c196b28befec6eefebba38e1R338 - so your concern shouldn't be that much of a problem.

But then you should filter before slicing here https://github.com/plone/mockup/pull/826/files#diff-b4d083c1c196b28befec6eefebba38e1R186

> However, adding the userid to the key, likewise as proposed for the fieldname shouldn't be a problem.
> What do you think?

I would add it, its probably easy and can save some headache...

thet:
jeez, what happens here... lots of test failures from other modules where I didn't expect them. First some interesting side effects due to mocking in plone.app.z3cform and Products.Archetypes, then http://jenkins.plone.org/job/pull-request-5.1/722/testReport/junit/plone.app/testing/layers_rst/ and http://jenkins.plone.org/job/pull-request-5.1/722/testReport/junit/plone.app.widgets.tests.test_utils/TestRelatedItemsOptions/test__base_relateditems_options/

Will investigate later.

PR #827, opened 2018-01-19
Add type=button to pickadate buttons to avoid form submit problems

cekk:
pickadate buttons (today and delete) didn't have a type="button" attribute.

If we use a date field with this widget inside a form, we need to have that attribute, to avoid these buttons listen to form submission events.

P.S.: the file has been saved with prettier, so it looks more different in the diff. If it's a problem, i'll revert it and commit only my changes without prettier.


thet:
LGTM.

PR #829, opened 2018-02-02
Modal performance: don't reposition after loading each img

cillianderoiste:
This reverts d06669e994f5a3:

   resize/position modal again once all images are loaded,
   plone/plone.app.toolbar#98

plone.app.toolbar is obsolete, and repositioning after loading each
image can get expensive. For example, if a modal contains a
DataGridField widget with reordering enabled and 20 rows, each row
contains 4 images, so the modal is repositioned 80 times.

The issue, plone/plone.app.toolbar#98, is no longer visible, so I'm
not sure if we run into other issues by removing this code.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,7 +14,8 @@ New features:
 
 Bug fixes:
 
-- add items here
+- Modal performance: don't reposition after loading each img
+  [cillianderoiste]
 
 
 2.7.0 (2018-01-27)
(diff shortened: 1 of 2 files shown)


frapell:
@cillianderoiste Where are you seeing issue plone.app.toolbar#98? I go to the repo and it doesn't seem to have "Issues" enabled in Github at least...

Have you tested how the modals behave when they have images? Do you have some steps to reproduce in order to test and validate?

If images break something when rendering the modal, maybe instead of calling ```positionModal()``` on every single image load, you can have some flag to be set, so after all images are loaded, you can call the method only once... just throwing an idea out there, not sure if it is possible...

thet:
IMO this looks ok.
Let's merge and see what happens! :)

PR #831, opened 2018-02-27
Make it possible to customize the available image classes

davisagli:
This adds an option to the tinymce pattern to make it possible to configure the available image classes / alignment options. I want to use this on one site to provide a "full width image" option.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,7 +10,9 @@ Breaking changes:
 
 New features:
 
-- add items here
+- Add `imageClasses` option to the tinymce pattern
+  to allow overriding the list of alignment classes.
+  [davisagli]
 
 Bug fixes:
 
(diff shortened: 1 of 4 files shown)


frapell:
LGTM :)

davisagli:
Thanks for merging. Now I need to make a PR for CMFPlone to update the plone-loggedin, right? Or is that part of the release process?

frapell:
@davisagli Yeah, doc is a bit hard to find, but AFAIK, you can follow https://github.com/plone/Products.CMFPlone/blob/master/DEVELOPING_BUNDLES.rst

thet:
@davisagli creating a bundle is not yet part of the release process. as soon as this is finished (it's almost-done) - https://github.com/plone/Products.CMFPlone/issues/1653 - it will get a bit easier.

the process is:
- create a bundle with ``plone-create-bundle``,
- update the last_compilation timestamp for the bundle which was changed in CMFPlone/profiles/dependencies/registry.xml
- make the obligatory changelog entry (one "Update mockup" per release is enough)
- And - if not done yet - create a upgrade step which re-imports the last_compiled timestamp.

All of that isn't really documented.

PR #764, opened 2017-05-19
Include Tinymce 4.6.1

frapell:

PR #832, opened 2018-02-28
Include TinyMCE 4.7.6

frapell:


frapell:
https://github.com/plone/Products.CMFPlone/pull/2320

frapell:
mmm... tests pass fine in my local machine... I will need to check again tomorrow...

frapell:
The issue in #764 seems to have been because of a piece of code in tinymce.js as follows:

```
    getStyleSheetLoader(editor).loadAll(editor.contentCSS, function (_) {
      initEditor(editor);
    }, function (urls) {
      initEditor(editor);
    });
```

Which should call initEditor after it tries to load the `content_css`, however, it seems karma does not serve the css files, so tinymce doesn't find the `content.min.css`.
I modified the karma config so it would serve this css, and even though I am able to get to the resource manually if I go to its URL, for some reason tinymce still can't (From the tests).

Now, this `initEditor` does this:

```
  var initEditor = function (editor) {
    editor.bindPendingEventDelegates();
    editor.initialized = true;
    editor.fire('init');
    editor.focus(true);
    editor.nodeChanged({ initial: true });
    editor.execCallback('init_instance_callback', editor);
    autoFocus(editor);
  };
```

And it is being called fine when you load tinymce in the browser, but it is not when doing it from the tests (Because of the css issue I mentioned).

The problem with this, is that when .save() is called, it has this:

```
      ...
      if (!elm || !self.initialized || self.removed) {
        return;
      }
      ...
```

which is the reason on why the test was failing in #764 but when manually trying tinymce in the browser, everything worked fine.

In order to have the tests passing, I am doing this:

https://github.com/plone/mockup/pull/832/commits/04a8905e9fa1906aac6f20bcd2481a0ce9edb8c8

As it can be seen, I was expecting to only do it in the tinymce test, and this worked fine, however this affects other tests, like the textareamimetypeselector...

I ended up adding the initialized flag in the pattern itself, this doesn't seem to bring any unwanted side effect, nor prevents the `initEditor` to be called, so I think it is pretty safe...

I don't know if this is a proper fix, or if someone has a better idea?

@thet or @vangheem could you guys review?

thet:
LGTM for me too.

mockup/patterns/tinymce/less/pattern.tinymce.less
```diff
@@ -1,6 +1,6 @@
-@import (less) "@{bowerPath}/tinymce-builded/js/tinymce/skins/lightgray/skin.less";
-@import (less) "@{bowerPath}/tinymce-builded/js/tinymce/skins/lightgray/Content.Objects.less";
-@import (inline) "@{bowerPath}/tinymce-builded/js/tinymce/plugins/visualblocks/css/visualblocks.css";
+@import (inline) "@{bowerPath}tinymce-builded/js/tinymce/skins/lightgray/skin.min.css";
+@import (inline) "@{bowerPath}tinymce-builded/js/tinymce/skins/lightgray/content.min.css";
```
datakurre: @frapell Would you rememer any specific reason for using ``(inline)`` instead of ``(css)`` here? I found out webpack builds failing, because ``(inline)`` makes less-leader import context to be the folder of pattern and webpack cannot find images and fonts below the skin folder. I wonder how normal Plone works with that.

frapell: @datakurre Honestly not... if you change it to ```(css)``` works? If so, I would say we can change it...

Issue #833, opened 2018-02-28
RelatedItems: Remember the breadcrumbs for a selected item


frapell:
When you navigate to a location inside the related items widget, the breadcrumbs reflect the current location where you are at. Then, you select an item and save

Then, if you re-edit the content, the chosen item is already in the widget, however the breadcrumbs is now outdated.

When the widget is configured to allow only 1 selection, and there is already an item selected, the breadcrumbs should start at the folder holding that item.

PR #834, opened 2018-02-28
When an item is already selected, update the breadcrumbs accordingly

frapell:
Rleated to gh-833


thet:
Sorry - messed up the PR with the issue.

frapell:
Thank you :)

PR #835, opened 2018-03-26
Fix edit link

davisagli:
/@@edit specifies a browser view, which is not available for Archetypes content items.  /edit should work more generally.


jensens:
I suppose resource need to be rebuild in Products.CMFPlone in order to make it effective.

PR #836, opened 2018-03-28
Use full path for tinymce fonts. 

frapell:
plone/Products.CMFPlone#2359


frapell:
There is no need to update Changelog since this should have been part of "Include TinyMCE 4.7.6"

PR #837, opened 2018-04-12
Improve relateditems sorting

davisagli:
I think this provides better defaults for sorting relateditems results. When searching it sets no sort index so results will be ranked by relevance to the search terms, but when browsing it sorts by getObjPositionInParent so the order will match what is shown in the structure pattern.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,11 @@ New features:
 
 Bug fixes:
 
+- Improved default relateditems pattern sort.
+  It now defaults to search term ranking when searching
+  and folder position (getObjPositionInParent) when browsing.
+  [davisagli]
+
 - *add item here*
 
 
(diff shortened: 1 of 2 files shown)


davisagli:
Oh thanks, I lost track of this one.

CHANGES.rst
```diff
@@ -14,6 +14,11 @@ New features:
 
 Bug fixes:
 
+- Improved default relateditems pattern sort.
```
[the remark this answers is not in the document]
davisagli: I guess you can argue that, but I considered it a bug that search results weren't sorted by relevance and folder listings weren't sorted by folder order.

mockup/patterns/relateditems/pattern.js
```diff
@@ -252,11 +252,18 @@ define([
             v: this.options.rootPath + this.currentPath + (this.browsing ? '::1' : '')
           });
 
+          var sort_on = this.options.sortOn;
+          var sort_order = sort_on ? this.options.sortOrder : null;
+          if (this.browsing && sort_on === null) {
+            sort_on = 'getObjPositionInPath';
```
thet: I've overseen this. should read ``getObjPositionInParent`` instead ``getObjPositionInPath``.

davisagli: Gah, right! Thanks for fixing.

PR #841, opened 2018-05-02
relateditems: fix sort_on index

thet:


thet:
Fixes: https://github.com/plone/mockup/pull/837/files#diff-b4d083c1c196b28befec6eefebba38e1R258

PR #842, opened 2018-05-09
Mark external link with rel="noopener"

mamico:
For performance and security external link marked as target=_blank must be opened as rel="nopener".

More info here:
https://developers.google.com/web/tools/lighthouse/audits/noopener
https://mathiasbynens.github.io/rel-noopener/

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,7 +10,9 @@ Breaking changes:
 
 New features:
 
-- *add item here*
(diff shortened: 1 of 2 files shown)


thet:
didn't know about this liink type. the ``rel="noopener"`` is also standardized: https://html.spec.whatwg.org/multipage/links.html#link-type-noopener - so LGTM.

PR #845, opened 2018-05-22
Part of PLIP 2150: date/time picker changes

jensens:
see https://github.com/plone/Products.CMFPlone/issues/2150


jensens:
Outdated, we should switch to a ReactJS based solution. 
If you plan to work further on this please reopen and go on.

PR #847, opened 2018-05-23
Fix issue where webpack was unable to load TinyMCE bundle styles

datakurre:
because files with relative resources were imported as with (inline) instead of (css)

https://github.com/plone/mockup/pull/832#pullrequestreview-122690361


datakurre:
Temporary fix for webpack was to alias the failing imports

https://github.com/collective/plonetheme.webpacktemplate/pull/11/commits/5ce3118560036e54206df9d33afdb52908180d86

PR #849, opened 2018-05-25
Fix for loading country specific languages codes

lyralemos:
With this fix, country specific language codes are correctly loaded. Fixes plone/Products.CMFPlone#949

As language codes are obtained via lang property in the html tag, there is an inconsistency with Plone locales code base. For instance, Brazilian Portuguese lang attribute is pt-br, but in plone locales is referred to as pt_BR.

This code does a simple transformation on the language code so that correct translations are loaded.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -36,7 +36,8 @@ Bug fixes:
   make sure that 'Original' is prefilled in the scale drop down
   [frapell]
 
-- Fix for loading country specific language codes [lyralemos]
+- Fix for loading country specific language codes
+  [lyralemos]
 
 
 2.7.2 (2018-04-08)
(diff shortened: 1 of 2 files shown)


davilima6:
@lyralemos will sign and send over the Contributor Agreement today. Thanks! (also kudos to @claytonc who initially helped with the solution=)

lyralemos:
I've already signed the agreement.
Code was committed using a different e-mail which I just added to my GitHub account.

davilima6:
I'd squash all commits before merge.

thet:
LGTM

lyralemos:
I've managed to run the tests locally and realized were the problem is.
For some reason tests are falling if I don't check if currentLanguage is different that en-us.

Should I create a new PR? Or can I still commit changes here?

gforcada:
@lyralemos a new pull request would be best, thanks for having a quick look at it!!

Issue #848, opened 2018-05-25
Include column sorting for the structure pattern


frapell:
We used to have this feature in Plone, where you could sort the elements inside the table for the folder_contents view, without affecting the actual order of items inside the folder.

You can rearrange items in the folder_contents, but that affects the actual order of items.

I am thinking on integrating a solution in order to provide this feature back, and I came across 2 alternatives:

https://datatables.net/ and http://issues.wenzhixin.net.cn/bootstrap-table/

I already integrated datatables in a project, so I know it works well with Mockup, but I don't know if it is something that it would be accepted to be included with the project.

What do you guys think?

frapell:
@vangheem @thet What do you guys think?

thet:
Sounds good to me + I'm not against it.

AFAIK Datatables brings a lot of table rendering functionality including search, which would make parts of structure patterns code obsolete. I'm not sure if it's easy to integrate.

Before it's going to be a major rewrite I'd put efforts to a React based rewrite of the structure pattern (plone-react). My plan is to look into a React option at the Buschenschanksprint ( https://community.plone.org/t/buschenschanksprint-2018-11-6-15-6-2018/6154 ) including an alternative to the Related Items Widget (a frequent source of complaints from my customers).

/cc @tisto any opinions?

thet:
Absolutely. Go for it :)

PR #850, opened 2018-05-30
Frapell sort folder contents

frapell:
I included the https://datatables.net/ library as a new pattern ```pat-datatables```, and included it to be used in the structure, so columns are sortable. I also included it with the ```plone``` bundle, so this is also available to editors when manually creating a table using TinyMCE (Maybe we could add in the future, a style in the drop down for TinyMCE to make it more easy to use... out of the scope of this PR though)

In addition, I went ahead and solved a couple of issues that the structure pattern was having (clicking disabled buttons, and allowing reordering after filtering with a query), as well as adding a new "250" option to the pagination size to show more items.

This is coming from gh-848


frapell:
@thet @vangheem @jensens Once this PR is merged, we need to also merge https://github.com/plone/Products.CMFPlone/pull/2436 and https://github.com/plone/plone.app.upgrade/pull/163

frapell:
Totally. Once I know the functionality is accepted I plan to backport to 5.1 and 5.0 as well

frapell:
Thanks for the feedback, will work on them and notify you when done

thet:
@frapell I did a rebase - I hope it didn't go wrong.
Tests are failing. I can't remember if they passed before my rebase... do you, @frapell ?

frapell:
@thet Yeah, they were, I will try to fix them when I can have some time to work on your suggestions... I wasn't able these past days :(

thet:
@frapell - restored old pre-rebased version. Let's see if tests are green.

frapell:
@thet I merged master back into the branch and added the ability to reset the ordering back to default without the need to refresh the whole view.

As per your review points:

* There is no explicit option to reset the datatables ordering, so that one is able to manually reorder items again.
  - This is now fixed. I added the ability to include a button in the status message, and it shows up like this:
image
This is the cleanest way I found to allow "unsorting"

* When clicking in the paginator (e.g. 250) the order is reset but the " Can not order items while manually sorting a column" is still in place. (items can be reordered, though).
  - Fixed, now the status message will be removed when changing the view

* When sorting via datatables, the "Rearrange" button is still active and works. Which is actually fine - no reason to deactivate (except the reorder-warning-message makes one believe that it won't work)
  - Well, the "rearrange" would change all items in the folder, so, as you say, it shouldn't be disabled, however I changed the status message to read "Can not drag and drop items to reorder while manually sorting a column" so it would be more explicit

* Why didn't the "All" paginagtion button work?
  - Currently, the batching is done using "backbone" and I couldn't find a way of choosing "All" as option for it. So, my workaround was to provide a really big number while showing "All" as text. However, that didn't work, because there is a hard coded limit in https://github.com/plone/plone.app.content/blob/master/plone/app/content/browser/vocabulary.py#L36, which throws and exception "Exception: Max batch size is 500". So, to allow "All" as an option, it would require major refactoring, and I think it would require a new PR, out of the scope of this one, so will just provide an additional option for 250 items, which I think is reasonable, and maybe add an "All" option in the future.

frapell:
@thet All green :)

frapell:
@thet Do you want to review again? or with your approval, I can just go ahead and merge?

thet:
Merging done for you!

Issue #857, opened 2018-07-13
structure pattern: moving on the tree doesn't update all context informations


cekk:
### What I did:

- Go to a folder (let's say News folder)
- Switch to folder_contents view
- Go up one level clicking on folder_contents breadcrumbs
- Delete the folder (News) from folder_contents buttons
- Select "Welcome to Plone" page
- Click "copy"

### What I expect to happen:

- The <title> should be updated
- "Welcome to Plone" page will be copied without errors
- The buttons should be updated with the current context url

### What actually happened:

- Some parts of the page are correctly updated, but page <title> is still "News"
- An alert appears with a message like this: "operation url http://localhost:8080/Plone/news/@@fc-copy is not valid"

### What version of Plone/ Addons I am using:

buildout.coredev (Plone 5.1.4.dev0)

thet:
Will be fixed with:
- https://github.com/plone/plone.app.content/pull/166
- https://github.com/plone/mockup/pull/876
resp.
- https://github.com/plone/plone.app.content/pull/167

The structure pattern does path replacement on button urls when navigating through the structure. The URLs from the pattern configuration were wrong in the first place.

See: https://github.com/plone/mockup/blob/master/mockup/patterns/structure/js/views/app.js#L280

PR #858, opened 2018-07-13
Update button urls when we change context

cekk:
This pr tries to fix some of the problems found in #857 

This is a wip implementation.
I don't know if this is the right approach.
Especially when i try to update the current button urls.


thet:
@cekk - as mentioned in #857, this will be fixed with:

- https://github.com/plone/plone.app.content/pull/166
- https://github.com/plone/mockup/pull/876
resp.
- https://github.com/plone/plone.app.content/pull/167

PR #861, opened 2018-08-09
Do not fail if initializing with another language or external script

frapell:


frapell:
@vangheem @thet Hey guys, I need some help with this issue... When I included Tinymce 4.7.6, I had an issue with tests, which I explain in https://github.com/plone/mockup/pull/832#issuecomment-369113718

Adding the flag as part of the pattern is not a good idea and it is the reason behind this issue https://github.com/plone/Products.CMFPlone/issues/2425

Without this flag set as part of the init code of the pattern, TinyMCE works just fine in the browser, however the issue is with the tests

Do you have any suggestion?

vangheem:
@frapell I'm really sorry, I haven't been in that code for so long and haven't used it either. Go ahead and take ownership and do what needs to be done.

tdesvenain:
adding `tinymce.tiny.initialized = true;` "fixes" the test but i have no idea if this would hide an actual issue

```
    it('on form submit, save data to form', function() {
      var $container = $(
       '<form>' +
       '  <textarea class="pat-tinymce">' +
       '  </textarea>' +
       '</form>'
      ).appendTo('body');

      var $el = $container.find('textarea');
      var tinymce = new TinyMCE($el);
      tinymce.tiny.initialized = true;  // ! HERE !

      tinymce.tiny.setContent('<p>foobar</p>');
      $container.submit(function(e) {
        e.preventDefault();
      });
      $container.trigger('submit');

      expect($el.val()).to.equal('<p>foobar</p>');
    });

```

frapell:
@tdesvenain Yeah, if you read my explanation in https://github.com/plone/mockup/pull/832#issuecomment-369113718 you can see at the bottom I mentioned about adding such a line in the tests, however that would only work when you call the pattern manually, ie. ```new TinyMCE($el);``` however, there are several other tests where the "pat-tinymce" class is used and then ```registry.scan($el);``` is called.

After banging my head for several hours around this, I wouldn't like to manually add ```tinymce.tiny.initialized = true; ```... That flag should be set as part of the tinymce init code, and as far as I understand, it has to do with the fact that tinymce is created asynchronously. I think the proper way around this would be for the tinymce tests to properly wait until the widget is created... I don't know how to do that.

This is why it works fine in the browser, but not in the tests...

frapell:
@sunew @thet @erral Can you guys give me a little help here?

Forcing the ```self.tiny.initialized = true;``` was a mistake I included because tests were not passing, however I realized afterwards, that it caused issues when TinyMCE was in any other language than English, or when loading external scripts...

However, removing that, causes test to not pass...

As I explained in these other comments, the failing of the tests (I believe) has to do with tinymce not being able to load ```/bower_components/tinymce-builded/js/tinymce/skins/lightgray/content.min.css```

In order to try and fix that I have made the changes from my last 2 commits

However now, the tests fail with "Some of your tests did a full page reload!" and I cannot figure out where or why...

sunew:
@frapell Interesting! I was looking into this yesterday with no luck - trying to play with init_instance_callback. 

(Like described here: 
https://www.tiny.cloud/docs-3x/customization/TinyMCE3x@Creating_unit_tests/
and here
https://gist.github.com/vojtajina/7400775 ) 
( but I had an issue with coordinating this with the require.js using the same __karma__.start method here mockup/tests/config.js )

Your approach looks good - did you try to change the 'on form submit, save data to form'-test to use the same 
`var tiny = createTinymce({`
as the other tests in mockup/tests/pattern-tinymce-test.js ?

frapell:
@sunew No, I did not try that... also, notice you are looking at doc for TinyMCE 3... is it still valid on 4 ?

sunew:
@frapell doh - trying to find similar for 4

sunew:
@frapell "Some of your tests did a full page reload!" fixed in #869 

The error was caused by another test, and must have been hidden by some other factors until your fixes. But the error was surfacing on master when using the FF or Chrome testrunners locally - this was the hint for me.

sunew:
A better way (instead of `tinymce.tiny.initialized = true;` ) would be to integrate tiny with karma.__start__ or similar; and to make the tests async with done() etc. So that tiny initializes correctly before the test is started.
I tried this but it would require some rewriting.
I suggest we merge #869 and then this one - the hard initializing only happens for the tests now, with @frapell 's fix, so it should have no influence on the actual code.

And we should test a migration from 5.1rc1 - @tdesvenain you had a setup and database for that already? Or is it already confirmed with your cherry-pick? (See https://github.com/plone/Products.CMFPlone/issues/2425 )

frapell:
@sunew Thank you A LOT for this... I think it is pretty clear you know a lot about JS testing, so feel free to implement your suggestions with karma and tinymce at any time ;)

sunew:
@frapell thanks! :)

ale-rt:
> @sunew Thank you A LOT for this... I think it is pretty clear you know a lot about JS testing, so feel free to implement your suggestions with karma and tinymce at any time ;)

+1, thanks a lot @sunew for the impressive job you are doing :)

PR #863, opened 2018-08-26
Structure fixes

thet:
Some changes which add to previous commits:

- The column count for the datatables initialization was wrong and led to a bunch of errors.

- Datatable CSS imports need to be with (inline) option, otherwise only a CSS import directive is included which leads to 404 errors.

- The Drag & Drop status message was not very clear to me. IMO thats better (@tkimnguyen plz have a look at that change)

Two new things:

- pat-structure: Use more tooltips in the actions menu.
- pat-structure: Move breadcrumbs out of folder contents table.

### DIFF
--- a/mockup/patterns/structure/js/views/table.js
+++ b/mockup/patterns/structure/js/views/table.js
@@ -147,7 +147,7 @@ define([
                     // Clear the status message
                     self.app.setStatus();
                   });
-        self.app.setStatus(_t('Cannot drag and drop items to reorder while manually sorting a column'), 'warning', btn = btn);
+        self.app.setStatus(_t('Drag and drop not possible while non-permanent column sorting active.'), 'warning', btn = btn);
         $(".pat-datatables tbody").find('tr').off("drag")
         self.$el.removeClass('order-support');
       } );
--- a/mockup/patterns/structure/js/views/table.js
+++ b/mockup/patterns/structure/js/views/table.js
@@ -76,7 +76,7 @@ define([
       var datatables_options = {
         "aaSorting": [],
         "aoColumnDefs": [
-          { "bSortable": false, "aTargets": [ 0, self.app.activeColumns.length + 1 ] }
+          { "bSortable": false, "aTargets": [ 0, self.app.activeColumns.length + 2 ] }
         ],
         "paging": false,
         "searching": false,
--- a/mockup/patterns/structure/js/views/table.js
+++ b/mockup/patterns/structure/js/views/table.js
@@ -71,12 +71,12 @@ define([
       var self = this;
 
       // By default do not start sorted by any column
-      // Ignore first column and the last one (activeColumns.length + 2)
+      // Ignore first column and the last one (activeColumns.length + 1)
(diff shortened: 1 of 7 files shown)


mockup/patterns/structure/js/views/table.js
```diff
@@ -147,7 +147,7 @@ define([
                     // Clear the status message
                     self.app.setStatus();
                   });
-        self.app.setStatus(_t('Cannot drag and drop items to reorder while manually sorting a column'), 'warning', btn = btn);
+        self.app.setStatus(_t('Drag and drop not possible while non-permanent column sorting active.'), 'warning', btn = btn);
```
thet: @tkimnguyen is that proper english and is the message clear enough?
My problem with the message before was, that it could have suggested that the "manual column sorting" was permanently stored.

sunew:
@thet could be good with a merge - can we get the tests to pass? I can perhaps help with guidance

mockup/patterns/structure/js/views/table.js
sunew: @thet 
"Notice: Drag and drop reordering is disabled when viewing the contents sorted by a column"
and then next to the message the button to reset the sort is shown.

sunew:
@thet looking at the tests - 3 out of 4 fixed.

I dont understand this change: "pat-structure: Use more tooltips in the actions menu." - What should be the effect here? We already have this:

selection_658

It causes a test failure that is hard to fix (for me ;) ).

sunew:
ok I get the tooltip change - was looking at the wrong place:)

selection_659

But it seems to have no effect? (I did a rebuild etc.)

sunew:
@thet "The column count for the datatables initialization was wrong and led to a bunch of errors.": I think @frapell is right - the change from 2 to 1 is wrong, as you can see from the screenshot here:

selection_659

What bunch of errors did you get?

sunew:
When i reverted back to 1, but kept the rest of the changes in the branch (and added an extra column)

selection_660

frapell:
@sunew I started the Jenkins tests... as soon as everything is green, we can merge, if @thet is done with his changes...

sunew:
@frapell I'm looking into building the plone bundles in a PR for cmfplone.
And i guess an upgrade step in pa.upgrade is needed, to update the bundle compilation dates, correct?
So all three would need to be merged 'together'.

frapell:
@sunew Yes, usually the order is:

1) Functionality is merged into mockup
2) Mockup is updated in CMFPlone
3) p.a.upgrade gets a new upgrade step for upgrading previous versions of CMFPlone

They don't need to be merged all together at the same time, but I usually leave a message linking all 3 PR so I'm sure they eventually get merged

sunew:
@frapell okay, thanks! I hope to have a cmfplone pr in a few hours

sunew:
@thet Have you got time to review the above? Especially the revert of "The column count for the datatables initialization" back to 2.

frapell:
@thet Any update on this? we have all green lights, should we merge it?

jensens:
LGTM. I'll merge.

PR #864, opened 2018-09-13
package-lock.json is supposed to be under revision control

sunew:
See https://github.com/npm/npm/blob/v5.0.0/doc/files/package-lock.json.md

"package-lock.json is automatically generated for any operations where npm modifies either the node_modules tree, or package.json. It describes the exact tree that was generated, such that subsequent installs are able to generate identical trees, regardless of intermediate dependency updates.

This file is intended to be committed into source repositories"

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -19,7 +19,9 @@ New features:
 - Add a new pattern, to provide the DataTables functionality
   https://datatables.net/
   [frapell]
-- *add item here*
+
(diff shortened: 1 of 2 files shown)


frapell:
I didn't know about this, and does sound like a very good idea. +1 from me

sunew:
@frapell I'll merge - this also avoids having it left behind as an artefact after a build.

Issue #675, opened 2016-05-21
Get rid of console deprecation messages: 


jensens:
In fact it does not hurt, but, three times ...

```
Usage of the mockup-patterns-base pattern is deprecated and it will eventually be removed.
Instead, use pat-base and explicitly set parser to 'mockup' when calling extend.
```

... and seven times ...

```
patterns.Patternslib Base: This pattern without a name attribute will not be registered!
```

... does not look very good for Plone core.

vangheem:
This is from patternslib. Maybe the issue should be moved there? Or, better yet I suppose. We should add names to those patterns or just refactor so they aren't real patterns.

sunew:
@vangheem @jensens 

It concerns:

LinkType
ExternalLink
InternalLink
UploadLink
ImageLink
EmailLink
AnchorLink

all from src/mockup/mockup/patterns/tinymce/js/links.js

We could give them a dummy name and trigger?

Originally name+trigger was removed in 

```
* 963f224a - remove name and trigger for LinkType--it is not really a pattern (2015-07-14) <vangheem>| 
| diff --git a/mockup/patterns/tinymce/js/links.js b/mockup/patterns/tinymce/js/links.js
| index f6bcc3ef..cf980169 100644
| --- a/mockup/patterns/tinymce/js/links.js
| +++ b/mockup/patterns/tinymce/js/links.js
| @@ -13,8 +13,6 @@ define([
|    'use strict';
|  
|    var LinkType = Base.extend({
| -    name: 'linktype',
| -    trigger: '.pat-linktype',
|      defaults: {
|        linkModal: null // required
|      },

```
If it does not hurt to reintroduce them, I can do it.

PR #868, opened 2018-09-16
Fix 675: Remove patternslib "This pattern without a name attribute will not be…

sunew:
… registered!" warnings by setting dummy name and trigger properties

Fixes https://github.com/plone/mockup/issues/675

If this approach is ok, merge.


sunew:
A rebuild of mockup for cmfplone is comming in another PR related to this: https://github.com/plone/mockup/pull/863
So I merge and close this one.

PR #873, opened 2018-09-27
Upgrade tinymce 

obct537:


mockup/patterns/tinymce/pattern.js
```diff
@@ -178,7 +177,7 @@ define([
       folderTypes: ['Folder', 'Plone Site'],
       tiny: {
         'content_css': '../../../bower_components/tinymce-builded/js/tinymce/skins/lightgray/content.min.css',
-        theme: '-modern',
+        theme: 'modern',
```
vangheem: I can't believe this was the problem...

PR #879, opened 2018-11-30
pat-resourceregistry: Add load_async and load_defer options.

thet:
- pat-resourceregistry:
  - Add input fiels for ``merge_with``, ``last_compilation``, ``jscompilation`` and ``csscompilation``.
  - Add ``load_async`` and ``load_defer`` options.

the resource registry looks now like: 
screenshot_2018-12-04 plone site 1

Merge:
https://github.com/plone/Products.CMFPlone/pull/2644
https://github.com/plone/mockup/pull/879
https://github.com/plone/plone.app.upgrade/pull/186

### DIFF
--- a/mockup/patterns/resourceregistry/js/fields.js
+++ b/mockup/patterns/resourceregistry/js/fields.js
@@ -366,6 +366,16 @@ define([
     }
   });
 
+
+  var MergeWithFieldView = ResourceSelectFieldView.extend({
+    multiple: false,
+    getSelectOptions: function() {
+      var self = this;
+      return ['', 'default', 'logged-in'];
+    }
+  });
+
+
   return {
     ResourceDisplayFieldView: ResourceDisplayFieldView,
     VariableFieldView: VariableFieldView,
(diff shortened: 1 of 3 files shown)


mockup/patterns/resourceregistry/js/fields.js
```diff
@@ -366,6 +366,16 @@ define([
     }
   });
 
+
+  var MergeWithFieldView = ResourceSelectFieldView.extend({
+    multiple: false,
+    getSelectOptions: function() {
+      var self = this;
+      return ['', 'default', 'logged-in'];
```
agitator: should this be a string field? or coming from a registry setting? thinking of other metabundles

thet: they are hardcoded anyways. see: https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/interfaces/resources.py#L119
I saved me the hassle to dynamize that which would involve making a proper vocabulary on the Plone site and so on...

PR #885, opened 2019-02-05
add navigationmarker pattern

agitator:


frapell:
@agitator I think to remember that it doesn't really matter, but shouldn't the name attribute be "navigationmarker" ? just for consistency...

agitator:
@frapell thanx, you're right
updated the name

PR #887, opened 2019-02-08
moment.js: lazy load locales

davilima6:
Fixes plone/Products.CMFPlone#1779.


davilima6:

mockup/patterns/moment/pattern.js
```diff
@@ -81,12 +81,15 @@
  */
 
 
+var lang = document.querySelector('html').lang;
+
 define([
   'jquery',
   'pat-base',
   'moment',
+  'moment-url/' + lang,
```
thet: I'd suspect this could be a problem in tests.
make sure. the lang attribute on html is available and provide a fallback like:
```
'moment-url/' + (lang || 'en')
```

davilima6: Unfortunately it wasn't enough. I pushed a couple more commits with that and also adding `moment-url` to `paths` in `config.js`

davilima6:

mockup/patterns/moment/pattern.js
davilima6: Indeed in tests <html> tag comes without `lang` attribute. In that case this suggests English as default, and since that's built-in in `moment.min.js` there's no locale to load.

We'd miss the test case for the lazy load actually working, i.e. checking for localised translations in case `<html lang="de">` is found in test mocks. What would the place to provide such `lang` attribute? In https://github.com/plone/mockup/blob/master/mockup/tests/fakeserver.js ? Could you give me pointers?

mockup/patterns/moment/pattern.js
```diff
@@ -146,5 +147,15 @@ define([
   });
 
   return Moment;
-
 });
+
+function lazyLoadMomentLocale() {
```
davilima6: @ale-rt , @agitator: oops, this function leaks to the global object. My goal was to hide some of the imperative code, so all patterns looked more similar between each other. I'll fix it in #890.

PR #889, opened 2019-02-10
Improve Related Items UX

davilima6:


MrTango:
Work is done, only the test has to be fixed, because of chnages in th selectors.

MrTango:
@davilima6 @thet I need some help with the test here, i can't run them locally because of problems with phantomjs not working. We changed some structure in the XML file, so the markup might be a bit different than before, i gues thats what is letting the test failing.

davilima6:
@MrTango , I checked all selectors from first failing case ('auto roundtrip') and they all seem to still exist. Also comparing with the templates I didn't detect any change that would affect the selectors as they are pretty simple, not much nesting or special selectors. Here's the expectation:
https://github.com/plone/mockup/blob/a0a8b0c9fb31e8f4e0ce9ac138cf8938ad80c319/mockup/tests/pattern-relateditems-test.js#L201

The error message is:

```
expected object with selector: '.pattern-relateditems-result-select' to have a length of 5 but got 0
```

But in the new templates that's still a class from a link element with data-path so I'm not sure why it's not finding it. Will look further in the weekend.

mockup/patterns/relateditems/pattern.js
```diff
@@ -217,7 +217,9 @@ define([
       // let's give all the options possible to the template generation
       var options = $.extend(true, {}, self.options, item, {
         'browsing': self.browsing,
-        'open_folder': _t('Open folder')
+        'open_folder': _t('Open folder'),
+        'urrent_directory': _t('current directory:'),
```
davilima6: typo here, missing C ;P

mockup/patterns/relateditems/pattern.js
```diff
@@ -378,6 +381,12 @@ define([
 
       $('.dropdown-toggle', self.$toolbar).dropdown();
 
+      // unbind mouseup event from select2 to override the behavior:
+      $(".pattern-relateditems-dropdown").unbind("mouseup");
```
davilima6: 👍 nice

MrTango: Yes, but we still need to customize select2, because of the keyboard handling. This is a bit to much to override. Otherwise we would have to copy over a big function to our code. But i made a fork of select2 and set the default branch there to 3.x, so that we can keep track of our customizations.

MrTango:
I think we should merge the branch in plone.staticresources first, because this brings the patches there.
https://github.com/plone/plone.staticresources/pull/16

mockup/patterns/relateditems/pattern.js
davilima6: Much better approach, I like it.

mockup/patterns/relateditems/pattern.js
```diff
@@ -286,12 +293,13 @@ define([
...
           // While browsing: always include folderish items
-          // Browsing and searching: Only include selectable items, which are not already selected.
+          // Browsing and searching: Only include selectable items which are not already selected, and all folders
+          // even if they're selected, as we need them available for browsing/selecting their children
           results = results.filter(
             function (item) {
               if (
                 (this.browsing && item.is_folderish) ||
-                (this.isSelectable(item) && this.selectedUIDs.indexOf(item.UID) == -1)
+                (this.isSelectable(item) && !this.selectedUIDs.includes(item.UID))
```
davilima6: oops, `includes` doesn't work for IE11 without a polyfill. I'll check if we have on in place otherwise I'll revert these changes.

davilima6:
Maybe it's easier to merge mockup first so that PR on static resources can already point to master. Also because if we merge resources first, then somebody else updates mockup (probably together with p.staticresources) we'll need a very similar PR to plone/plone.staticresources#16 just to update resources again.

That's the approach we took with Lazy Locales PR for Moment.js. We postponed resources merge until mockup tests passed and in fact we're still pending to merge the PR in resources, because we want to try to improve the functional mockup solution, already in master.

So as a rule of thumb it seems to me we might save ourselves some work if we always merge mockup first, then p.staticresources. Does this make sense? I might be missing some scenarios...

davilima6:
🎉

mockup/patterns/relateditems/pattern.js
davilima6: oops, think we merged it with a typo, @MrTango

PR #890, opened 2019-02-13
Momentjs: improve fallback for lazy load locales

davilima6:


davilima6:

mockup/patterns/moment/pattern.js
```diff
@@ -150,12 +149,29 @@ define([
...
   }
 
+  // Taken from https://github.com/moment/moment/blob/3147fbc486209f0b479dc0b29672d4c2ef39cf43/src/test/moment/format.js#L463-L468
+  var MOMENT_LOCALES =
+    'ar-sa ar-tn ar az be bg bn bo br bs ca cs cv cy da de-at de dv el ' +
+    'en-au en-ca en-gb en-ie en-nz eo es et eu fa fi fo fr-ca fr-ch fr fy ' +
+    'gd gl he hi hr hu hy-am id is it ja jv ka kk km ko lb lo lt lv me mk ml ' +
+    'mr ms-my ms my nb ne nl nn pl pt-br pt ro ru se si sk sl sq sr-cyrl ' +
+    'sr sv sw ta te th tl-ph tlh tr tzl tzm-latn tzm uk uz vi zh-cn zh-tw';
+
```
jensens: so, if _momentjs_ changes, the list needs an update, right? I fear theres no way to compile it automatically from somewhere?

davilima6: Correct. They shouldn't change much as Moment has 127 locales already so they _should_ be stable but yes, it will eventually need human labor this way.

Moment has API to list all _loaded_ locales so we can fetch automatically from `moment-with-locales.js` where they're loaded but AFAIK I cannot import it to this pattern file without defining it and therefore loading it in the browser, beating the purpose of the PR.

davilima6: Other alternative I see would be to pre-cook it as part of the build process (not sure if Python or Javascript ones, probably both), store in the registry and allow to be consumed as Javascript module or global -- I imagine a registry view that delivers json. Then new versions of the pattern could at least worry about updating this registry instead of a loose string here.

Any ideas, @thet ?

jensens: if its this complete I am fine to hardcode it here. It was just a question.

PR #894, opened 2019-02-16
Fix bugs on Moment.js lazy loading

davilima6:
- Avoid leaking lazyLoadMomentLocale to global namespace
- Fix bug by not lower casing mockup-i18n's currentLanguage
- Fix bug on IE11 due to lack of support for 'includes'

### DIFF
--- a/mockup/patterns/moment/pattern.js
+++ b/mockup/patterns/moment/pattern.js
@@ -96,22 +96,20 @@ define([
     'mr ms-my ms my nb ne nl nn pl pt-br pt ro ru se si sk sl sq sr-cyrl ' +
     'sr sv sw ta te th tl-ph tlh tr tzl tzm-latn tzm uk uz vi zh-cn zh-tw';
 
-  lazyLoadMomentLocale(currentLanguage);
-
   function isLangSupported(lang) {
     return MOMENT_LOCALES.split(' ').indexOf(lang) !== -1;
   }
 
-  function lazyLoadMomentLocale(lang) {
+  function lazyLoadMomentLocale() {
     var LANG_FALLBACK = 'en';
 
-    if (lang === LANG_FALLBACK) {
+    if (currentLanguage === LANG_FALLBACK) {
       // English locale is built-in, no need to load
       return;
     }
 
     // Format language as expect by Moment.js, neither POSIX (like TinyMCE) nor IETF
-    lang = lang.replace('_', '-').toLowerCase();
+    var lang = currentLanguage.replace('_', '-').toLowerCase();
 
     // Use language code as fallback, otherwise built-in English locale
     lang = isLangSupported(lang) ? lang : lang.split('-')[0];
@@ -122,6 +120,8 @@ define([
 
     require(['moment-url/' + lang]);
   }
+
+  lazyLoadMomentLocale();
 
   var Moment = Base.extend({
     name: 'moment',
--- a/mockup/patterns/moment/pattern.js
+++ b/mockup/patterns/moment/pattern.js
@@ -104,7 +104,8 @@ define([
     var LANG_FALLBACK = 'en';
 
     if (currentLanguage === LANG_FALLBACK) {
-      // English locale is built-in, no need to load
+      // English locale is built-in, no need to load, so let's exit early
+      // to avoid computing fallback, which happens at every loaded page
       return;
     }
 
--- a/mockup/patterns/moment/pattern.js
+++ b/mockup/patterns/moment/pattern.js
@@ -86,7 +86,42 @@ define([
   'mockup-i18n',
   'moment'
 ], function($, Base, i18n, moment) {
-  lazyLoadMomentLocale();
+  var currentLanguage = (new i18n()).currentLanguage;
+
+  // From https://github.com/moment/moment/blob/3147fbc/src/test/moment/format.js#L463-L468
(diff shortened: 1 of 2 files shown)


davilima6:

jensens:
Sometimes things are more complex than they seem, right? But now it really looks fine.

mockup/patterns/moment/pattern.js
```diff
@@ -86,7 +86,42 @@ define([
   'mockup-i18n',
   'moment'
 ], function($, Base, i18n, moment) {
-  lazyLoadMomentLocale();
+  var currentLanguage = (new i18n()).currentLanguage;
```
ale-rt: Not a merge blocker, but is this needed here?
I would just evaluate it in the lazyLoadMomentLocale function.

mockup/patterns/moment/pattern.js
```diff
@@ -86,7 +86,42 @@ define([
...
+
+  // From https://github.com/moment/moment/blob/3147fbc/src/test/moment/format.js#L463-L468
+  var MOMENT_LOCALES =
+    'ar-sa ar-tn ar az be bg bn bo br bs ca cs cv cy da de-at de dv el ' +
+    'en-au en-ca en-gb en-ie en-nz eo es et eu fa fi fo fr-ca fr-ch fr fy ' +
+    'gd gl he hi hr hu hy-am id is it ja jv ka kk km ko lb lo lt lv me mk ml ' +
+    'mr ms-my ms my nb ne nl nn pl pt-br pt ro ru se si sk sl sq sr-cyrl ' +
+    'sr sv sw ta te th tl-ph tlh tr tzl tzm-latn tzm uk uz vi zh-cn zh-tw';
+
+  lazyLoadMomentLocale(currentLanguage);
```
ale-rt: I would move this invocation after the function is defined and would call it without an argument (currentLanguage can be evaluated in the scope of lazyLoadMomentLocale, see comment above).
Not a merge blocker.

mockup/patterns/moment/pattern.js
```diff
@@ -86,7 +86,42 @@ define([
...
+
+  function isLangSupported(lang) {
+    return MOMENT_LOCALES.split(' ').indexOf(lang) !== -1;
+  }
+
+  function lazyLoadMomentLocale(lang) {
+    var LANG_FALLBACK = 'en';
```
ale-rt: There is something fishy in this function.
You pass the lang variable and then reevaluate it (twice).
By just reading it I feel puzzled by the code.
Anyway not a merge blocker, we can improve this later.

mockup/patterns/moment/pattern.js
davilima6: I moved this same line from a few lines below so I didn't have to repeat myself.

Also I preferred to fetch from mockup-i18n than directly from document.querySelector('html') so that if we change the implementation there, or if we want some transformations, Moment/js pattern get directly from there.

ale-rt: Oh, I see that this is also used later in another method.

mockup/patterns/moment/pattern.js
davilima6: You're right. I'll change that.

About the function declaration since it's hoisted should be fine but I'll double check, probably your suggestion makes things more readable.

mockup/patterns/moment/pattern.js
davilima6: I'll see what I can improve. I put 2 early exit conditions because I didn't want to apply transforms and checks to the original `lang` string if it's `en`, it's minor but I tried to avoid because this code is currently executed at every page load.

mockup/patterns/moment/pattern.js
davilima6: I pushed the 2 suggested changes but I must confess I wasn't too sure about the 2nd one. It seems to me not relying on local scope would be safer, for instance the function could become easier to test and also move around, for example, if we wanted it to go in a `moment-locale` pattern (i don't like bloating the pattern file, in this case it was OK-ish because it's little code but I'd prefer if it looked more similar to the others: define wrapper + extend Base + return extended object. yet I preferred to tackle that in a next PR)

mockup/patterns/moment/pattern.js
davilima6: I don't pass `lang` anymore but I'm not sure how to improve more. Now that I don't reassign a received param, do you think it's fine to reassign it twice?

I could replace the ternary by plain ifs but I found it more verbose and less logical as the two first parts of the ternary repeat.

ale-rt: If that is causing much pain than value ignore my comment. 
It is not a merge blocker and it was just suggesting bad feelings :)
Really if this is ready please merge it!

davilima6:

davilima6:
@ale-rt, no pain! Just asking for suggestions on how to improve it. I'll merge then, thank you!

ale-rt:
Test failure unrelated, I am merging this one

mockup/patterns/moment/pattern.js
davilima6: Your opinion is very welcome so I'm glad to remove code smells! I appreciate it!

davilima6: just noticed var LANG_FALLBACK should live outside function. If there are any other pending changes I improve that :)

PR #901, opened 2019-03-01
Upgrade to TinyMCE 5.0.1

frapell:


frapell:

vincentfretin:
Like @erral said in https://community.plone.org/t/list-of-tinymce-defects-in-plone-5-2-call-for-fixing-issues-and-participation/12257/14
we don't use bower anymore so we may now use directly the npm version of tinymce instead of using the repo https://github.com/artursmirnov/tinymce_builded ?

erral:
I don't know the implications of changing to use the npm-based release of tinymce, but we will find a more updated version for sure and will be easier to handle than relying on updating the bower-based tinymce_builded package.

frapell:
I don't know why the process of including tinymce was by using that "builded" package instead of just using the official release. I agree with @erral, I guess we can try to pull from the official release and see how it goes

thet:
@erral @frapell that was because the official release on github (which was our main source of packages when we used bower) did not have any translations included.
That's probably different with releases from npm, so I'm happy to switch to npm releases and drop tinymce-builded.

erral:
It looks like installing tinymce with npm doesn't download the translations.

thet:
Closed in favor of: https://github.com/plone/mockup/pull/1024

Issue #903, opened 2019-05-23
Inlinevalidation pattern: Should not submit files


frapell:
The current implementation of the inline validation pattern would submit the whole form when performing validation on each field

https://github.com/plone/mockup/blob/d75be61cd2e4eb6f323d8281b151c40be4a28d30/mockup/patterns/inlinevalidation/pattern.js#L127-L137

Which means that when there's a field (or even worse, more than 1) with a big file, it would require the whole file (or files) to be submitted to the server after each inline validation.

Steps to reproduce:
1) Add a File content type
2) Choose a big file from your computer for the file field
3) Click the title field, enter some text and select another field so the inline validation would trigger

With the "Network" tab opened in the browser's dev tools, you can see the request to ``@@z3cform_validate_field`` taking really long to finish.

If working in localhost, you can play with the "throttling" feature to limit the upload speed in order to make it more evident.

PR #904, opened 2019-05-23
pat-inlinevalidation: Avoid submitting files

frapell:
The way I am fixing this issue, is by 
1) clone the original form
2) remove any ``<input type="file">`` from it
3) submit this cloned form instead

If an error comes from the validation, it will be rendered in the original form

### DIFF
--- a/mockup/patterns/inlinevalidation/pattern.js
+++ b/mockup/patterns/inlinevalidation/pattern.js
@@ -98,11 +98,14 @@ define([
         var $input = $(input),
             $field = $input.closest('.field'),
             $form = $field.closest('form'),
+            $cloned_form = $form.clone(),
             fname = $field.attr('data-fieldname');
 
+        // XXX: Remove binary files so they are not uploaded to server
+        $cloned_form.find("input[type=file]").remove();
         this.queue($.proxy(function(next) {
-            $form.ajaxSubmit({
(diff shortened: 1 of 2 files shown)


frapell:

frapell:
@ale-rt Mmm, I did not find where or how the exception through CSS class happens... and if you see my description of the issue in https://github.com/plone/mockup/issues/903 the way this works is by submitting the whole form, and then in the backend, it would check from all the errors, and only return the one for the field that triggered the inline validation
https://github.com/plone/plone.app.z3cform/blob/28be7cd91f317e9cfba75fd3dd48753a9402e442/plone/app/z3cform/inline_validation.py#L57-L60

What you mention about fields like the captcha is a very good point... I believe the reason that today "works" is because usually people would mark the checkbox when they finish filling the form... if they would mark the captcha and then fill the form, then the issue of it being "consumed" already will arise when the form is submitted...

I don't know what a proper solution could be, because now that I am thinking about it, there are at least 3 scenarios:

1) Fields that should have inline validation
2) Fields that shold not be submitted when doing validation on other fields (Like the binary files)
3) Fields that should have inline validation but should only be submitted when validating them, or not with their actual values (Say you want to validate that someone entered the catptcha but you do not want to yet validate the captcha itself)

I will merge this one for now...

PR #909, opened 2019-06-06
Fixes Products.CMFPlone#2490

fulv:


fulv:
Fixes Products.CMFPlone#2490

davilima6:
Hey, @fulv, thanks for another fix! I tested and `z-index: 4` is enough for lateral toolbar but `2` is needed for top one. Could you please confirm? Here's the panel for changing it http://localhost:8080/Plone/@@site-controlpanel

I think it's OK to merge with this strategy as it's urgent to get it into a release.

But maybe another approach would be to leave `.navbar { z-index: 10 }` as it was and instead bring the toolbar up: increase the `z-index` of the `#edit-zone` (Toolbar pattern), which is [currently](https://github.com/plone/mockup/blob/4b696bad8feb99e9e259cde94c30245a7c3fff57/mockup/patterns/toolbar/pattern.toolbar.less#L10) `3` ([overwrote to](https://github.com/plone/mockup/blob/4b696bad8feb99e9e259cde94c30245a7c3fff57/mockup/patterns/toolbar/pattern.toolbar.less#L466) `5` for `.plone-toolbar-left`) to something like `1000` to give room for global UI widgets. The only thing it comes to my mind one would want above toolbar would be modals or dialogs, specially in case they result from clicking toolbar's buttons. What do you think?

Thinking a step further, and possibly out of scope for this fix, but if we had `position: relative; z-index: 10` on `.container` we'd get an independent stacking for anything inside it ("local content widgets") so patterns or 3rd-parties like embeds can have any z-index (say, 999999999) and still won't cover toolbar. I couldn't think of drawbacks, yet :)

fulv:
Thanks for your review, @davilima6!
I decided to set the z-index on the edit bar to a more conservative 20.
Your other suggestion, to change `.container` to `relative`, broke a bunch of stuff, so I left it alone.

fulv:
@Davi Lima <davilima6@gmail.com> regarding the .content relative, I had it
with the toolbar on top and it was completely unclickable, i.e. there was
something transparent with a higher z-index on top of it.  Plus there was a
link showing in the top left that is normally invisible, I can't remember
what it was.

Thanks for merging!
-- 
Fulvio

PR #913, opened 2019-06-24
Upgrade testing dependencies: Karma and Mocha

davilima6:
Updated versions:

- [x] Karma: from ~3.0.0 to ~4.1.0
- [x] Mocha: from ~3.5.3 to ~6.1.4
- [x] Chai: from ~3.5.0 to ~4.2.0

<hr>

I postponed `karma-sauce-launcher` for now because it [failed installing](https://travis-ci.org/plone/mockup/jobs/549814555) `selenium-webdriver@4.0.0-alpha.4` with "incompatible linux platform" for `fsversion@1.2.9`: 
- [ ] ~karma-sauce-launcher: from 1.2.0 to 2.0.2~ 

```
yarn install v1.3.2
[1/5] Validating package.json...
[2/5] Resolving packages...
[3/5] Fetching packages...
info fsevents@1.2.9: The platform "linux" is incompatible with this module.
info "fsevents@1.2.9" is an optional dependency and failed compatibility check. Excluding it from installation.
error selenium-webdriver@4.0.0-alpha.4: The engine "node" is incompatible with this module. Expected version ">= 10.15.0".
error Found incompatible module
```

<hr>

Finally we may include Sinon too if we solve the problem reported [below.](https://github.com/plone/mockup/pull/913#issuecomment-505065684)

- [ ] Sinon: from ~1.17.1 to ~7.3.6 (#914)

PR #914, opened 2019-06-24
Upgrade sinon

davilima6:
WIP. Currently failing with:

```
  Structure traverse links
    ✗ test navigate to folder push states
	undefined is not an object (evaluating 'history.pushed.url')
	tests/pattern-structure-test.js:1306:28

    ✗ test navigate to folder push states - urlStructure
	undefined is not an object (evaluating 'history.pushed.url') A
	tests/pattern-structure-test.js:2153:28

    ✗ test navigate to folder push states - pushStateUrl
	undefined is not an object (evaluating 'history.pushed.url')
	tests/pattern-structure-test.js:2190:28
```

See: https://github.com/plone/mockup/pull/913#issuecomment-505065684

### DIFF
--- a/mockup/js/config.js
+++ b/mockup/js/config.js
@@ -3,7 +3,7 @@
 
 /* global module:true */
 
-(function() {
+(function () {
   'use strict';
 
   var tinymcePlugins = [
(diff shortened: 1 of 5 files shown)


davilima6:
I also tried replacing the custom Sandbox by the default one available since Sinon 5.0 but same errors persisted.

thet:
Green! merging.

davilima6:
How did you make it pass? :D It was not a flaky failure, was it?

thet:
I did some test fixes but appended to your last commit. So not easy to demonstrate :)

I renamed the ``history`` method to ``history_stub`` to not accidentally cross the global window.history.
Also, instead of ``sinon.sandbox.stub`` I used ``sinon.stub`` directly and passed the ``history_stub`` object as value instead as callFake.

Something like that.

davilima6:
Got it, nice! I hadn't noticed the changes, thought you pushed force only to update from master. Thanks!

PR #915, opened 2019-06-25
Upgrade jQuery from version 1.11.3 to 1.12.4

davilima6:


davilima6:

davilima6:
Thanks! Yes, I can do it but it may take me some days as I just went into vacation.

davilima6:
I'm trying to figure out which bundle to recompile after this jQuery upgrade.

@agitator, @thet: How can I compile `plone-base`, where jQuery seems to live currently? (btw it seems to me we should remove its requirejs definition from [plone-logged-in.js](https://github.com/plone/plone.staticresources/blob/cfa2ee677c692ce9423ff12885c5fbc614c353fc/src/plone/staticresources/static/plone-logged-in.js#L17-L22), would u agree?)

`bin/plone-compile-resources --bundle=plone-base` fails after this: `Warning: Task "compile-plone-base" not found. Use --force to continue.` Is it because of something missing from plone/plone.staticresources#27?

davilima6:
Indeed it seems we miss the Grunt task for compiling `plone-base`: running `bin/plone-compile-resources` without specifying any bundle won't generate `plone-base-compiled.js`. Also I just noticed it won't generate `plone-legacy-compiled.js` either so I wonder how these two should be built.

davilima6:
@thet, which bundle needs to be recompiled? I see `jquery.min.js` is delivered standalone (not sure why: is it because it's registered as `stub_js_modules`?)

davilima6:
I noticed the `jquery.min.js` (that is delivered standalone by Plone) seems unrelated to the jQuery upgrade in this PR. So I created plone/plone.staticresources#34 which is from where that file is actually served. @thet, could you please confirm this is how it should go?

thet:
@davilima6 sorry for late answer - I'm currently in vacation myself.
But yes, jquery is delivered standalone, see:
https://github.com/plone/Products.CMFPlone/blob/bd98bd05a3eaf1c5b9579a7fb49da9c37f6721fd/Products/CMFPlone/resources/browser/combine.py#L78
https://github.com/plone/Products.CMFPlone/blob/bd98bd05a3eaf1c5b9579a7fb49da9c37f6721fd/Products/CMFPlone/resources/browser/scripts.py#L101

You should be able to build the plone-base bundle via ``./bin/plone-compile-resources -b plone-base`` (note: if you use ``-G`` you're skipping gruntfile generation).
If it doesn't work - did you run all upgrade steps? Maybe try on a fresh Plone installation. It works for me...

davilima6:

davilima6:
Thanks, @thet! I definitively need to spend more time in this `resources` module.

I managed to build `plone-base` but with the confirmation that resource `plone.resources/jquery.js` is served separately from bundles - and that the version updated in this PR is not even the one from that resource (btw where is the one in this PR used?) - it seems to me there's nothing to build on `plone.staticresources` after this PR. Is that the case?

Also it seems to me an upgrade step is not needed. To get new jQuery one needs only to upgrade `plone.staticresources`, which will already drop in the new vendored jQuery in the same file path.

With these two confirmations I think we can merge this one and then #916.

---

To succeed running `./bin/plone-compile resources -b plone-base` I had to discard my `Plone` site and create a new. But because I had all upgrade steps applied (and never ran with -G, did not know it, thanks!) I'm afraid something may be missing from the upgrade steps until 5.2 final. I'll investigate further but fyi site was originally created in step `5205` which I then upgraded to `5208`.

davilima6:
Thanks, @jensens. Now we need to update complementary plone/plone.staticresources#34 before merging it.

Then I think we can merge #916 next. I saw you updated its branch so I just triggered Jenkins on it.

Please let me know if any related task requires attention. Cheers!

Issue #921, opened 2019-09-13
Tinymce pattern: Link popup looses tab selection on active link type


parruc:
When I insert an external link in tiny selecting some text and clicking the link button, the resulting popup has several tabs including internal, upload, external ecc.

When I inserted one or more links, save the content and edit it back, when I try to change the link the selected tab is not the one where I entered the data before but it's always "internal". That's bad cause I have to click on all tabs to find what kind of link it was.

mamico:
#922 merged

PR #922, opened 2019-09-16
Fix Tinymce pattern: Link popup looses tab selection on active linktype

mamico:
see #921


mamico:

PR #930, opened 2019-12-20
Backport lost tinymce translations

djowett-ftw:
There might be more work to do here, but I would like to backport this translations fix
 
However ... I need pointers of what to do next


djowett-ftw:

djowett-ftw:

mockup/patterns/tinymce/js/links.js
```diff
@@ -466,6 +466,14 @@ define([
         upload: this.options.upload,
         text: this.options.text,
         insertHeading: this.options.text.insertHeading,
+        insertImageHelp: this.options.text.insertImageHelp,
+        ExternalImage: this.options.text.ExternalImage,
```
thet: is ``externalImage`` in mockup master

thet: there are also other upper/lowercase inconsitencies here.
@djowett-ftw could you please just copy over the whole generateModalHtml function from master? I'd also like to have the attributes in the same order.

thet:
@djowett-ftw please fix the upper/lower case issues and update the templates as well.
For which Plone version do you need these changes?

After this gets merged you need to create a new bundle for plone-logged-in for CMFPlone (no staticresources available for Plone < 5.2) with ./bin/plone-compile-resources

mockup/patterns/tinymce/js/links.js
djowett-ftw: Hi @thet - good idea I've done that.

djowett-ftw:
> @djowett-ftw please fix the upper/lower case issues and update the templates as well.
> For which Plone version do you need these changes?

Hi @thet - thanks for the guidance, sorry it took a while.   I think this is done now, as you requested.  We are looking for this to work with Plone 5.1 (we're currently on 5.1.6).

There are some missing German translations I'll put in plone.app.locales which depend on this too.

> After this gets merged you need to create a new bundle for plone-logged-in for CMFPlone (no staticresources available for Plone < 5.2) with ./bin/plone-compile-resources

Understood

djowett-ftw:

djowett-ftw:
@thet I think this is correct now - with https://github.com/collective/plone.app.locales/pull/287
But the test above hasn't reported back even though it appears to be finished.

djowett-ftw:
Trying to close / open to get Jenkins to restart

thet:
@djowett-ftw tnx for taking care of this!

Issue #932, opened 2020-01-02
Need to escape characters when using the query filter


frapell:
In the structure pattern, when entering text in the search input, we need to escape special characters.

As an example, if you use `&` as part of the text, for instance `Text1 & Text2` the query is truncated and the backend receives
```
(Pdb) query                                                                                                                                                                   
'{"criteria":[{"i":"SearchableText","o":"plone.app.querystring.operation.string.contains","v":"Text1'
```

This of course causes issues, since it is not valid JSON.

PR #933, opened 2020-01-02
Structure pattern: Escape special characters when filtering

frapell:
This fixes gh-932

### DIFF
--- a/mockup/patterns/structure/js/views/textfilter.js
+++ b/mockup/patterns/structure/js/views/textfilter.js
@@ -93,7 +93,7 @@ define([
         clearTimeout(self.timeoutId);
       }
       self.timeoutId = setTimeout(function() {
-        self.term = $(event.currentTarget).val();
+        self.term = encodeURIComponent($(event.currentTarget).val());
         self.app.collection.currentPage = 1;
         self.app.collection.pager();
 
--- a/news/932.bugfix
+++ b/news/932.bugfix
@@ -1,0 +1,2 @@
+Structure pattern: Escape special characters when filtering
+[frapell]


frapell:

mockup/patterns/structure/js/views/textfilter.js
```diff
@@ -93,7 +93,7 @@ define([
         clearTimeout(self.timeoutId);
       }
       self.timeoutId = setTimeout(function() {
-        self.term = $(event.currentTarget).val();
+        self.term = encodeURIComponent($(event.currentTarget).val());
```
davilima6: this turns spaces into %20, for instance, and is more used to encode get params. does the filter decode already?

alternatively we could transform 'Text1 & Text2' into `Text1 &amp; Text2` with `new Option($(event.currentTarget).val()).innerHTML` _(maybe `event.currentTarget.value`, without jquery, works even in ie11)_. but still the filter would need to decode these.

frapell: @davilima6 I honestly don't mind any solution... I used encodeURIComponent because it works and encodes everything properly... For instance, I submit the following text in the filter query:

`This is my text including special characters , / ? : @ & = + $ # and some ét®æß`

I then see this under the network tab 
image

And I see this in the backend
```
(Pdb) query
{'criteria': [{'i': 'SearchableText', 'o': 'plone.app.querystring.operation.string.contains', 'v': 'This is my text including special characters , / ? : @ & = + $ # and some 
ét®æß*'}, {'i': 'path', 'o': 'plone.app.querystring.operation.string.path', 'v': '/::1'}], 'sort_on': 'getObjPositionInParent', 'sort_order': 'ascending'}
```

(This is with a pdb in line 636 of `plone.app.vocabularies.catalog`)

If your solution works like this, then I'm happy :)

davilima6: If the filter already parses encoded URIs then no need for change, thanks for the fix

Issue #678, opened 2016-05-30
Structure Query Popover: broken location criteria


davilima6:
I failed trying to reproduce [this observation](https://github.com/plone/mockup/issues/381#issuecomment-43166686) from @djay, and actually noticed the behavior seems quite broken: https://dl.dropboxusercontent.com/u/5766075/screencasts/broken-location-criteria.mp4

Should we remove Location for 5.1 final while this isn't sorted out? It might take me a while to get familiar with this (Backbone) codebase.

djay:
can’t seem to get your mp4 to play. can you run it through gifbrewery or similar so you can paste in in github? or write out the problem?

davilima6:
Hi, @djay, thx for warning. The video's encoded in H264 so you need gstreamer1.0-libav in Ubuntu. I guess you're on OSX so I tested it and could watch the mp4 in VLC. Nevertheless here's the original file, this one should work ootb in browser:

djay:
@davilima6 H264 shouldn't be an issue on osx but obviously some codec is having problems. You are right, VLC plays it fine. I've found it really useful to learn an animated gif too however. An issue is a record of the problem so it should be self-contained.
Gifbrewery doesn't seem to like webm either so I'd have to transcode it first :(

So looks like the issue is that the query function is ignoring relative path criteria (but not absolute paths). Do you know if the catalog is receiving the relative path criteria? Perhaps the context of the request is set to the root?

davilima6:
Doesn't Gifbrewery open the mp4? Anyway VLC is really quick at transcoding, just Cmd+Shift+S.

And, no, I didn't have the time to debug it yet. Waiting for status:confirmed first ;-)

djay:
yes, I worked it out. gif uploaded now (not sure what went wrong with the colours). I didn’t realise gif brewery is paid. I must have got it in the first 2 weeks by accident when it was free. There are plenty of others though, inc online converters.

davilima6:
Gif brewery _really_ didn't like my mp4 so I found this site which did a better job: http://ezgif.com/video-to-gif

btw I also updated a video I uploaded yesterday to: https://github.com/plone/mockup/issues/671#issuecomment-222504072

thet:
@davilima6 @djay the location query filter seems a bit useless where the filter is applied only on the current view.
:/

jensens:
[...]

Issue #672, opened 2016-05-17
Structure alert isn't removed after query filters are cleared


davilima6:
Steps to reproduce
1. Add a criteria, e.g. `Title`
2. Notice the warning alert: `Can not order items while querying`
3. Remove all criteria
4. Notice the warning doesn't go away

newbazz:
I would like to work on this issue.

thet:
Fixed with: https://github.com/plone/mockup/pull/937

Issue #671, opened 2016-05-17
structure not intuitive how to query across folders (via improved structure filter ui/ux)


davilima6:
# Implementation

Trying  to solve #381 I propose:
1. **Initial value for Title criteria:** When user inputs something in Live Filter and then clicks the `Query` button, Plone should open the Query popover with initial `Title` criteria filled with that input. This makes it explicit why and how the current listing is filtered already and why user won't see all results even if he/she clears all the popover filters.
2. **Different buttons for Live filter and Query popover :** using `glyphicon-search` suggests users to think that button is needed to be pressed after providing input in the text field, which 1) isn't true, as the filtering happens by keystroke and 2) leads to unexpected ui behavior as a popover opens (instead of searching/filtering/querying). So I suggest:
   1. Update placeholder from "Filter" to "Filter items" to make label paralel to search viewlet's placeholder
   2. Add a button with `.glyphicon-filter` or `.glyphicon-option-horizontal` for the Advanced Query popover 
   3. Move `.glyphicon-search` inside and to the right of the input field, to reinforce this is a search input widget while still keeping it visually simple with only one explicit button (besides the search one would still work when js is disabled so no a11n loss)
   4. Add styles to new Query button to indicate query state, i.e. whether any criteria other than Title (which is already visible in Live filter) is set. For the relative path criteria, state would be empty when different than current folder's path.
3. **In-page live filter** should return items **_containing_ querystring**, instead of **only starting with it.**, as in  plone/plone.app.users#54.
4. Add a `in current folder` checkbox (ticked by default) beneath live filter. Besides, add `Location` criteria by default on the `More filters` popover. When checkbox is on, `Location` is set to `Current (./)`. When it's off, `Location` value is default, that is, `Custom`, which provides a nice navigation widget initially set at `entire site` level. This way users gets expected behavior (wider search) and at the same time the chance to refine their query.
# Risks
- Too many buttons/clutter?

davilima6:
If there's consensus I can work on the tasks above.

@vangheem @djay @jensens @thet what do u think?

jensens:
All sounds good to me. 

Only thing I'd like to see additional is a kind of "Reset Query" feature (button, broom-icon, ...).

djay:
Cool but #381 had two options, which one are you implementing?

 Perhaps you need mockup to show how all these ideas are going to look. Its a bit  hard to follow.

[...]

Issue #381, opened 2014-05-15
structure not intuitive how to query across folders


djay:
It would be hugely productive to use structure to be able to do adhoc searches across a whole site for example, find all unpublished files and publish them.

Not sure the best way to achieve this but one idea is this
- currently clicking the query button lets you filter only the current folder. It effectively is like a collection query with an implicit relative path criteria. We could make this relative path criteria explicit. It would be there by default but you can remove it if you want. This has the bonus of teaching people about how to use a relative path criteria.
- another option might be to have simple checkbox of "search only current folder". This checkbox could explitly add or remove the relative path query from the query builder?

jensens:
Closing as outdated. Mockup was completely rewritten for Plone 6 (jQuery/Backbone/RequireJS → ES6/Patternslib). Old structure cross-folder query.
If you think this issue is still relevant in the new implementation, please reopen it. Thank you for the original report!

PR #937, opened 2020-02-03
Structure pattern: Filter UI improvements, etc.

thet:
Structure pattern:
- Show statusmessage with "Clear filter" button when filters are applied.
- Show visually if a filter is set.
- Allow multiple status messages.
- Display toolbar a bit compacter.

- Instead of clearing the filter when changing the path (initial approach from first commit) now there is a status message shown with a "Clear filter" button in it. For this to work it was necessary to allow multiple status messages. The "Clear filter" status message must always be present while other status messages must still be possible.
That way it should be very obvious that the current view is filtered.
- If a text filter is set, the filter input has a blue border.
- if a querystring is set, the magnifier icon has a blue background.
- The status messages look a bit nicer.

Applied filter:
Screenshot from 2020-02-03 02-01-54

New statusmessage design + more compact toolbar (no border-bottom, no margin-bottom, no min height):
Screenshot from 2020-02-03 02-00-09

Old statusmessage design:
Screenshot from 2020-02-03 01-50-18

Merge together:
https://github.com/plone/mockup/pull/937
https://github.com/collective/plone.app.locales/pull/286

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - 8.12.0
+  - 12.14.1
 env:
   global:
     - secure: "fPRKgxcbSXyiLqkUjAjPkn5eGBqMqf6c5haTZ9x76sDl1792nDyiL2FXSpiv2s8PeOqMbWgcXly0r+CLXypYkWba9nIxb4bhvtXyDoYAaoBV0lG1NbBakuOwqW/u60WpPJS1qTqmEUSEbSxdntqk7HF1yG6MdMi3PdubxMyw9h0="
--- a/mockup/Gruntfile.js
+++ b/mockup/Gruntfile.js
@@ -158,7 +158,14 @@ module.exports = function(grunt) {
       console.log('read folder: ' + path);
       var files = fs.readdirSync(path);
       files.forEach(function(filename) {
-        if (filename === 'node_modules' || filename === 'node_modules') {
+        if (
+          filename === '__pycache__' ||
+          filename === 'build' ||
+          filename === 'coverage' ||
+          filename === 'less' ||
+          filename === 'node_modules' ||
+          filename === 'tests'
+        ) {
           return;
         }
         var stats = fs.statSync(path + filename);
--- a/mockup/js/ui/views/button.js
+++ b/mockup/js/ui/views/button.js
@@ -35,7 +35,7 @@ define([
         }
         _.each(this.extraClasses, function(klass){
           this.$el.addClass(klass);
-        });
+        }.bind(this));
 
         if (this.tooltip !== null) {
 
--- a/mockup/js/utils.js
+++ b/mockup/js/utils.js
@@ -354,6 +354,13 @@ define([
     }
   };
 
+  var createElementFromHTML = function(htmlString) {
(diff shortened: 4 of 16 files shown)


agitator:
Learning a new feature in Plone everday ... ;-)

My users use the text search feature to navigate quickly within folder structures with lots of content.

Never expected a feature like that, though I think it's neat, I also have my doubts that many users know and use it.

To make this feature useful imho a status message about activated filters is not helping much. I'd rather each query paramater as we do in a collection - but I think that should not be the scope of this improvement.

- I'd still say if a user filters the listing with a simple text entry, that entry should be reseted if one changes the location/context via breadcrumbs.
- If we really want to improve the querystring feature, we would have to add a collection like experience and add a query string element for searchable text (and remove) the text entry from that search field as soon as one adds further search query. This way it would be clearer what exactly happens

thet:
I try so summarize the main question here:

Initially the idea was to clear the text filter as soon as the path changes. That is actually super simple and done in this commit: https://github.com/plone/mockup/pull/937/commits/cd19401381603d480e17007bbed9c2e7371339e2

After implementing this I found that I also have to clear the query filter. But just clearing all filters without a chance for the user to prevent seemed to harsh.
1) Query are not so easily built (many clicks necessary)
2) It could be quite useful to keep the query while browsing (e.g. filter for Images and Folders to browse for images).

To avoid another button like "Keep filter while browsing" I implemented it so that any filters/queries are kept while browsing but showing a "Clear Filter" button in a notification.

/cc @davilima6 @frapell @jensens @djay any input on this?

Referencing other Issues related to this PR here - for reference. I dont want to expand the scope of this PR any further:
https://github.com/plone/mockup/issues/381
https://github.com/plone/mockup/issues/671
https://github.com/plone/mockup/issues/672
https://github.com/plone/mockup/issues/678

davilima6:
I like the described solution, will try to give it a go soon. Thanks, @thet!

davilima6:

thet:
@davilima6 I don't understand - the notification in light-blue and the buttons when filter are applied in green? I'm absolutely open for color changes.
I noticed some flaws with the notifications. Currently the "drag/drop reorder" stacks upon multiple times. Have to rethink the notification system there.

davilima6:
Agree, we should check if notification is already in list before appending it.

And to your question, yes. IMO:
* Only show the yellow one with exclamation mark to real warnings like `Cannot order item while querying`
* The one for "Some filters applied..." should be an info level only and have light blue background. I'd _try_ to avoid the red background also in the `Clear filters` button, maybe have a secondary background color, maybe white will give enough contrast with the blue background
* Success ones like "Successfully moved item" should be in green

thet:
@davilima6 off topic: are you coming to the alpine city sprint? I plan to work on patternslib/mockup there: https://docs.google.com/document/d/172UtDjJbIfnoxHE_x67KkirOfnhcRJ1casBZlyjJGe0/edit#heading=h.7hs2swxej1x1

davilima6:
@thet, unfortunately not this year :~ I'm taking vacations in homeland in the period 🇧🇷 But Salutations to All! Hope to be next year.

thet:
@davilima6 enjoy the time!

PR #920, opened 2019-09-11
TinyMCE: fix positioning of sub-menus in a modal

cillianderoiste:
This uses the ui_container configuration option in TinyMCE to anchor the position of menus to the parent modal.

https://github.com/tinymce/tinymce/commit/7fd61eeac61e9a88dfe8ed3c4c0398b5ee0328e2

Closes #867

PR #945, opened 2020-02-26
Generate a random id and append it to the container_id

fredvd:
because there might be more than one TinyMCE in the DOM.


fredvd:

fredvd:
This Improves #920

PR #962, opened 2020-04-02
show and hide actionmenu elements depending selection and copied elem…

### LINKED ISSUE #961: Structure pattern: action buttons should only be present if there is some action to do
As part or an accessibility review we were requested to hide all actions present in the folder contents view (copy, paste, cut, rearrange, change properties, workflow, ...) if there is no items selected. Morever, the paste action should only be visible is there is something to paste.

@ionlizarazu

erral:
…ents

Fixes #961 

@ionlizarazu

### DIFF
--- a/mockup/patterns/structure/js/actions.js
+++ b/mockup/patterns/structure/js/actions.js
@@ -94,6 +94,14 @@ define([
       var self = this;
       e.preventDefault();
       self.doAction('copy', _t('Copied'), _t('copying'));
+      //check if something has been copied
+      if (!$.cookie('__cp')) {
+        //if there is nothing copied don't show paste button
+        $("#btngroup-mainbuttons #btn-paste").hide()
+      }else{
+        //if there something copied show paste button
+        $("#btngroup-mainbuttons #btn-paste").show()
+      }
     },
     pasteClicked: function(e) {
       var self = this;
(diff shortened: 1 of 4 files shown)


erral:

jensens:
I am not sure if hiding is the best solution, or if it is better to grey out/ disable the buttons.

Opinions?

erral:
Yeah, it should be something that does not allow to make any action when clicking on them, hiding is one option.

jensens:
@thet @agitator  - any opinions on this?

thet:
I think disabling would be less disturbing instead of hiding/showing depending on the state. It probably also makes it harder to orientate in the UI when buttons disappear on different places.
On the other side hiding cleans up the UI.
If disabling the buttons is also an option accessibility wise, then I'd go for disabling.

There are already buttons disabled (should be at least), including when pasting and nothing was selected:
Button initialization: https://github.com/plone/mockup/blob/65b92f2178df68013ccccf7bdd4354add870422d/mockup/patterns/structure/js/views/app.js#L357
Paste button disable/undisable: https://github.com/plone/mockup/blob/65b92f2178df68013ccccf7bdd4354add870422d/mockup/patterns/structure/js/views/app.js#L236
ButtonGroup: https://github.com/plone/mockup/blob/65b92f2178df68013ccccf7bdd4354add870422d/mockup/js/ui/views/buttongroup.js
ButtonView: https://github.com/plone/mockup/blob/65b92f2178df68013ccccf7bdd4354add870422d/mockup/js/ui/views/button.js

vincentfretin:
This was a regression. See my fixes in https://github.com/plone/mockup/pull/996

PR #949, opened 2020-03-23
Do not use lists where items are not lists

erral:
Fixes #948

PR #970, opened 2020-04-03
do not mix different type of list items in the same list and add styl…

### LINKED ISSUE #967: Structure pattern: pagination uses a list item to show just an informative message
Here: https://github.com/plone/mockup/blob/master/mockup/patterns/structure/templates/paging.xml#L30

This "Show" is inside the list but it's not a list item, it's there just to have the same look-and-feel of the pagination items.

2020-04-02 14-38-53-ko-pantaila-argazkia

erral:
…ing to the new markup

The styling was wrong after merging #949, now we have fixed it here because it was also needed here.

See:

2020-04-03 14-25-12-ko-pantaila-argazkia

Fixes #967 

@ionlizarazu

### DIFF
--- a/mockup/patterns/structure/less/pattern.structure.less
+++ b/mockup/patterns/structure/less/pattern.structure.less
@@ -74,6 +74,14 @@
         border: 1px solid @btn-info-border;
     }
 
+    .pagination{
+        vertical-align: middle;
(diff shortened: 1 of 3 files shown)


erral:

PR #987, opened 2020-06-24
Fix regression with moment relative date not localized

vincentfretin:
See https://github.com/plone/Products.CMFPlone/issues/2953 for details

### DIFF
--- a/mockup/patterns/moment/pattern.js
+++ b/mockup/patterns/moment/pattern.js
@@ -83,10 +83,13 @@
 define([
   'jquery',
   'pat-base',
+  'pat-registry',
   'mockup-i18n',
   'moment'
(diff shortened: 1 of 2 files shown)


vincentfretin:
With the minified build only, I still had a race condition for the date columns in folder contents
https://github.com/plone/mockup/commit/5fde80900cd057ba009e24b14afd0309984e5161 fixed it.

vincentfretin:
And another changes I had to write https://github.com/plone/mockup/commit/af672aede8cb09f1983e6949df96bf7b855227d9 to avoid a warning, but I'm not sure why it triggered. Probably several bundles are launching the pat-moment pattern and there is a race condition. Anyway this commit fixes the warning.

PR #990, opened 2020-07-01
Fix CSS issue on "Pattern options" tab in Resource Registries

vincentfretin:
See https://github.com/plone/Products.CMFPlone/issues/1111

### DIFF
--- a/mockup/patterns/resourceregistry/pattern.resourceregistry.less
+++ b/mockup/patterns/resourceregistry/pattern.resourceregistry.less
@@ -90,16 +90,11 @@
     }
 
     .lessvariables,
-    .patternoptionsuse {
+    .patternoptions {
         .form{
             padding-top: 15px;
             clear: both;
         }
-    }
-
-    .pat-resourceregistry .lessvariables .form {
-        padding-top: 10px;
-        clear: both;
     }
 
     .patternoptions {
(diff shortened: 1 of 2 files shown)


mockup/patterns/resourceregistry/pattern.resourceregistry.less
```diff
@@ -90,18 +90,13 @@
     }
 
     .lessvariables,
-    .patternoptionsuse {
+    .patternoptions {
```
vincentfretin, line 93: This was clearly a typo. patternoptionsuse doesn't exist anywhere in the code.

mockup/patterns/resourceregistry/pattern.resourceregistry.less
```diff
@@ -90,18 +90,13 @@
...
             padding-top: 15px;
             clear: both;
         }
     }
 
-    .pat-resourceregistry .lessvariables .form {
-        padding-top: 10px;
-        clear: both;
-    }
-
```
vincentfretin, line 104: This rule doesn't make any sense. We are in `.pat-resourceregistry` already so the final rule is `.pat-resourceregistry .pat-resourceregistry .lessvariables .form` so it's never used. The rule that is used is the one just above with padding-top 15px.

vincentfretin:

vincentfretin:
If you want to test, in Resource Registries, check "Development Mode(only logged in users)", press "Develop CSS" on "resourceregistry" bundle, click "Save. Then refresh the page (F5) and go to "Pattern Options".
The fix will be in a Plone release version when we will rebuild the bundles in plone.staticresources with latest mockup master.

PR #999, opened 2020-07-07
[WIP] Structure: added hidden badge like expired

ekulos:


vincentfretin:
Hi, thanks for the contribution.
Can you please tell us what you are fixing? What are the steps to test this? Can you create a screenshot to show us what it looks like?
You will need to add a changelog entry by creating a file in the news folder, `news/999.feature`, you can see the other files as example.

thet:
I like the feature @ekulos but you need the plone contributor agreement: https://docs.plone.org/develop/coredev/docs/contributors_agreement_explained.html

Also, maybe hidden is not the best ``term`` for content which is excluded from navigation? On the other hand, it's short, which is good.

I was thinking adding this feature myself, so we do not have to wait for @ekulos to sign the contributor agreement....

What do you think @vincentfretin ?

vincentfretin:
Go ahead! :)

petschki:
successor -> #1176

PR #1005, opened 2020-07-30
respect default_page_types in actionmenu if folder_contents

pbauer:
Fixes https://github.com/plone/Products.CMFPlone/issues/2700
Together with https://github.com/plone/plone.app.content/pull/208

### DIFF
--- a/mockup/patterns/structure/js/actionmenu.js
+++ b/mockup/patterns/structure/js/actionmenu.js
@@ -108,7 +108,7 @@ define(['underscore', 'translate'], function(_, _t) {
       delete result['move-top'];
       delete result['move-bottom'];
     }
-    if (model.is_folderish || !app.setDefaultPageUrl) {
+    if (app.defaultPageTypes.indexOf(model.portal_type) == -1 || !app.setDefaultPageUrl) {
       delete result['set-default-page'];
     }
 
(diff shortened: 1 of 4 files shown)


pbauer:

pbauer:
I'm not sure why travis fails with: 
```
undefined is not an object (evaluating 'app.defaultPageTypes.indexOf')
```
Could it be that the build requires the source-checkout of plone.app.content or did I miss something when writing the code? It worked fine locally.

Also: Once this is merged `plone.staticresources` needs to be added to checkouts.cfg of coredev? Do I need to do all the other things it says in https://github.com/plone/plone.staticresources/#how-to-upgrade-the-resources-in-this-package?

vincentfretin:
The mockup tests doesn't depends on Plone, do a search for the failing test "use special view action for special types". The test is defined in `mockup/tests/pattern-structure-test.js`. You may need to add `defaultPageTypes` to the mock defined there.
Travis actually seems to test only with PhantomJS 8 times. :) Please test with the `make test` command (PhantomJS) and then with `make test-dev` (Chrome) locally. Both are green in master.

Yes you need to do all the things described in https://github.com/plone/plone.staticresources/#how-to-upgrade-the-resources-in-this-package and you need to checkout plone.staticresources yes like the documentation says. I updated it recently when I did the last upgrade step, I hope the instructions are clear.

pbauer:
@vincentfretin thanks for the info. I fixed the tests and will now proceed to follow your docs 😄

PR #1017, opened 2020-10-09
[WIP] pat-structure: badge for exclude_from_nav

thet:
pat-structure: Add a hidden badge for content which is excluded from the navigation.

Same as https://github.com/plone/mockup/pull/999
but with me as author who has signed the contributor agreement.

Not for merge now.
Also want to reconsider if "hidden" is a good term for this + maybe add some "title" attributes to further explain the badges.


petschki:
successor -> #1176

PR #1024, opened 2020-10-09
Upgrade TinyMCE to 5.5.1.

thet:
Replaces: https://github.com/plone/mockup/pull/901

Currently doesn't load because troubles with the theme.

I want to switch to the official TinyMCE release and not use ``tinymce_builded``. All translations are managed on Transifex.
Our options for getting the translations is a) Download them, or b) Use a CDN for them. The tiny cloud CDN might cost something, so that leaves us with downloading the translation files.

Later on I want to dyamically ``import()`` them.

More info:
tiny cloud CDN: https://github.com/tinymce/tinymce/issues/2352#issuecomment-619523008
transifex: https://www.transifex.com/tiny/tinymce/
config: https://www.tiny.cloud/docs/configure/localization/#language
download: https://www.tiny.cloud/get-tiny/language-packages/
npm: https://www.npmjs.com/package/tinymce
current tinymce problems: https://community.plone.org/t/list-of-tinymce-defects-in-plone-5-2-call-for-fixing-issues-and-participation/12257/16

If there is anyone who want to take over, please go ahead!


thet:
TinyMCE 5 will be included in Plone 6 via the Mockup ES6 effort:
https://github.com/plone/mockup/pull/1025

PR #1027, opened 2020-10-20
Make bootstrap tabs in Resource Registries

balavec:


balavec:
I made updates so this is now working for both Plone 6 and Plone 5.2.

thet (approved):
LGTM

PR #1028, opened 2020-10-23
Fix toolbar being updated only on the first navigation action

fredvd:
Main issue: this fixes the links in the plone toolbar becoming outdated when one navigates more than once in in the folder contents view. 

But wait, there's more...

What still is very strange and I think unwanted is that the toolbar updating is done twice.  And it's not only the toolbar, for every navigation click in folder_contents, 3 ajax calls are made twice to the backend for:
* `@@render-toolbar`  (toolbar itself)
* `@@fc-contextInfo` (actions in the toolbar)
* `@@getVocabulary`   (folder_contents main table)

### Double event handlers? ###

When I inspect the registered event handlers on the body element in the Firefox devtools, the toolbar structure-changed event is present twice , once from jquery and once from DOM2|bubbling, but I have no clue if this causes the double refresh of the toolbar. 
folder_contents first load events on body

Also: another click event handler in the Toolbar component is not properly unregistered before it's reregistered and pile up on the body tag when navigating :

https://github.com/plone/mockup/blob/590d229599433bac4f2265ce7dfa7857c408bf17/mockup/patterns/toolbar/pattern.js#L247

Changing this line 247 to::

       $('body').off('click').on('click', function(event) {

Solves the click events not being cleaned up, but this is rather new for me, I'm already happy I kind of found the toolbar update isssue. :-)  If somebody more knowledgable could confirm this is the case?

### DIFF
--- a/mockup/patterns/toolbar/pattern.js
+++ b/mockup/patterns/toolbar/pattern.js
@@ -514,10 +514,12 @@ define([
           $.ajax({
             url: $('body').attr('data-portal-url') + path + '/@@render-toolbar',
           }).done(function(data) {
-            var $el = $(utils.parseBodyTag(data));
-            $el = $el.find('#edit-zone').length ? $el.find('#edit-zone') : $el;
-            that.$el.replaceWith($el);
-            Registry.scan($el);
+            var $newel = $(utils.parseBodyTag(data));
+            var hasedit = $newel.find('#edit-zone').length;
+            $newel = hasedit ? $newel.find('#edit-zone') : $newel;
+            var $replacetoolbar = $('#edit-bar').find('#edit-zone');
+            $replacetoolbar.replaceWith($newel);
+            Registry.scan($newel);
           });
         });
 
--- a/mockup/patterns/toolbar/pattern.js
+++ b/mockup/patterns/toolbar/pattern.js
@@ -244,7 +244,7 @@ define([
           }
         });
 
-      $('body').off('click').on('click', function(event) {
+      $('body').off('click.closetoolbarsubmenus').on('click.closetoolbarsubmenus', function(event) {
         var $el = that.$container.find(event.target);
         // we need to check if the target isn't the nav which can be
         // triggered if we click on the portal-header and plone-toolbar-more-subset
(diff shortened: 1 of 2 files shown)


mauritsvanrees (approved):
This looks like a good change, and it works for me. I would like another review though.

I initially did not see a difference. I use current coredev 5.2 Python 3.8, with your mockup branch checked out. Fresh Plone Site. Set development mode On in resource registries, save it. That *should* be enough I think.
But after `./bin/plone-compile-resources`, and turning development mode Off again, it works.

Scenario:

- Create Fresh Plone Site.
- Create private Folder.
- Go to folder contents.
- In folder contents go to the site root. You see the toolbar change, which worked already before, since this is the first click.
- In the same folder contents go to the private folder. You see the toolbar change again. Most visibly: the review status changes. This is the part that did not work before.

mauritsvanrees:
Had contact with Fred. After putting resource registries in development mode, I have to develop the `plone-editor-tools` and then it works.

mockup/patterns/toolbar/pattern.js
```diff
@@ -244,7 +244,7 @@ define([
           }
         });
 
-      $('body').on('click', function(event) {
+      $('body').off('click').on('click', function(event) {
```
thet, line 247: This would unregister all events handlers bound to the body on "click" event. Modal dialogs closing on a click outside the dialogue could be a case.
To prevent this, you could factor out the inline callback into a normal function which you pass also the ``off`` method as second argument: https://api.jquery.com/off/

thet:
@fredvd I does look right except for the one comment but I'm currently short on time to reproduce the failure fully.
I can confirm the doubled requests, which shouldn't happen.
After fixing the issue I mentioned, go ahead and merge this.

mockup/patterns/toolbar/pattern.js
thet: Sorry - I've got something wrong: You'd have to pass the callback as a third argument.
Instead do this:
```suggestion
      $('body').off('click', 'FRED-PLEASE-FIND-A-NAME').on('click', 'FRED-PLEASE-FIND-A-NAME', function(event) {
```
please replace 'FRED-PLEASE-FIND-A-NAME'  with a suitable name - quickly looking at the code I don't know what it does.

fredvd: @thet thanks! I've updated the .off().on() calls.  To prevent removing all clicks you can add a .namespace value to the event name itself.  I have changed the generic one with 'click.closetoolbarsubmenus'.  It closes opened submenu's when an editor clicks outside the toolbar in the main content.

There are a few other places where we don't use namespaced but maybe should, I'll propose changed to thosein a separate pull request.  First get this one fixed & released.

mauritsvanrees:
Tested (in the compiled version from https://github.com/plone/plone.staticresources/pull/105) and it works.

Should be done on master as well.

PR #1025, opened 2020-10-10
ES6

thet:
[ES6 refactoring, using webpack and babel. Built upon the MrTango/11tydocs branch. Depends on latest ES6yfied patternslib. Following the refactoring in https://github.com/patternslib/Patterns

```
yarn install
yarn build
yarn build:stats
npx webpack-bundle-analyzer stats.json
npx jest src/pat/autotoc
```

Current status:
Migrated patterns are uncommented in src/patterns.js. Currently there are autodoc, backdrop, contentloader and cookietrigger. A bundle can be built but no tests are migrated and no demos are tried in the browser.

### Patterns

Put your name behind the pattern name to indicate that you are owrking on this pattern!

Status:
functional | tested | repsoncible
- [x] [x] autotoc /thet
- [x] [x] backdrop /thet
- [x] [x] contentloader /thet
- [x] [x] cookietrigger /thet
- [x] [x] datatables /MrTango
- [x] [x] eventedit  /thet -> Removed. Was unused. A script in plone.app.event does the job currently.
- [x] [ ] ~~filemanager~~ >> not used anymore in core, will not be migrated by us /MrTango
- [x] [ ] formautofocus /fredvd /jan
- [x] [x] formunloadalert /jmevissen
- [ ] [ ] inlinevalidation /thet
- [x] [ ] livesearch /veritazx
- [x] [x] markspeciallinks /yuri
- [x] [x] modal /Gomez /petschki #1068
- [x] [x] moment /thet -> Replaced by pat-display-time from Patternslib
- [x] [x] navigationmarker
- [x] [x] passwordstrength /veritazx, yuri
- [x] [x] ~~pickadate~~  >>  pat-datetime-picker (patterslib)
- [x] [x] preventdoublesubmit
- [x] [x] querystring /petschki
- [x] [ ] recurrence /thet
- [x] [ ] relateditems /MrTango
- [x] [x] select2 /thet
- [x] [ ] sortable /talarias
- [x] [ ] structure
- [x] [ ] textareamimetypeselector
- [x] [ ] texteditor /thet  >> we will integrate: pat-code-editor (patternslib)
- [x] [ ] ~~thememapper  >> needs update, editor only, but we will give up the Diazo rules builder~~
- [x] [ ] tinymce / MrTango #1065 
- [x] [x] toggle /Gomez (https://github.com/plone/mockup/pull/1034)
- [x] [x] toolbar /Gomez /agitator
[...]

PR #1034, opened 2020-12-12
Migrate toggle to es6

Gomez:


thet:
Merged into: https://github.com/plone/mockup/pull/1025

PR #1039, opened 2021-02-17
Fix Folder contents default page marker and upload context. 

fredvd:
Don't remove the eventlistener on this scope before re-adding, the others are then also removed.

### DIFF
--- a/mockup/patterns/structure/js/views/table.js
+++ b/mockup/patterns/structure/js/views/table.js
@@ -31,8 +31,8 @@ define([
       self.subsetIds = [];
       self.contextInfo = null;
 
-      $('body').off('context-info-loaded').on('context-info-loaded', function(event, data) {
-
+      $('body').on('context-info-loaded', function(event, data) {
+        
         self.contextInfo = data;
         /* set default page info */
         self.setContextInfo();
--- a/mockup/patterns/structure/pattern-structureupdater.js
+++ b/mockup/patterns/structure/pattern-structureupdater.js
@@ -26,7 +26,7 @@ define([
 
     init: function() {
 
-      $('body').off('context-info-loaded').on('context-info-loaded', function (e, data) {
+      $('body').on('context-info-loaded', function (e, data) {
 
         if (this.options.titleSelector) {
             $(this.options.titleSelector, this.$el).html(data.object && data.object.Title || '&nbsp;');
--- a/news/3220.bugfix
+++ b/news/3220.bugfix
@@ -1,0 +1,2 @@
+Fix Folder contents default page marker and upload context.
+[fredvd]


mauritsvanrees:
Problem is how to review this.

When I use the 3.x branch, put the resource registries in development mode, and click Develop Javascript for the plone-logged-in bundle, it already works. So this is without your changes.
With your changes, it also works.

Then I reset the resource registry changes, so we are back in production mode. I run steps 1-4 as outlined in the [`plone.staticresources` README](https://github.com/plone/plone.staticresources/blob/1.x/README.rst#how-to-upgrade-the-resources-in-this-package). I skip steps 5-7 for creating an upgrade step, but that should be fine: I create a fresh Plone Site after this. I check the folder contents in this new site: still no marker visible for the default front-page.

Here is my branch in case you want to try with a fresh Plone Site:
https://github.com/plone/plone.staticresources/tree/maurits/build-20210217
It is definitely possible that I am doing something wrong.

BTW, when running `make test` in your mockup branch, I get 300 succeeded tests, and one failure:
"TinyMCE test reopen add link modal FAILED"

When I do the same in branch 3.x, I have no failures:

```
PhantomJS 2.1.1 (Mac OS X 0.0.0): Executed 301 of 304 (skipped 3) SUCCESS (9.626 secs / 8.799 secs)
TOTAL: 301 SUCCESS
```

(In both cases I have to press `CTRL-C` at the end.)

O, never mind: when I retry on your branch, the tests pass...

thet:
@fredvd @mauritsvanrees I encountered the same problem (wondering why this wasn't an issue until now). I'll take care of this PR!

mauritsvanrees:
@thet I looked at it with Fred. Problem why with his PR it was working for him and not for me, was because the real change is in the plone-editor-tools bundle and not in the plone-logged-in bundle that I built on my plone.staticresources branch.

mauritsvanrees:
@thet But maybe you have a different solution in mind in `mockup`?

fredvd:
@thet  When I 'fixed' the context update problems in the toolbar a few months ago I noticed that the body tag accumulated events when navigating in the folder contents.  In several places. Hence I added the `.off()`.  This is fine as long there is one trigger and one listener. 

But in case of 'context-info-loaded' there are 3 listeners:   the toolbar add menu, the title/description updater and setContextInfo for pathbar and default-page. 

The nasty part: when you 'develop javascript' for plone-editor-tools, the problem seems to go away, also with the current Plone 5.2.3 bundled static resources. But with a compiled bundle  `.off('context-info-loaded').on('context-info-loaded') I think the .off()  removes one or both of the other 2 listeners.

So you can only test the issue and fix by running plone-compile-resources to update the bundle on filesystem or building the bundle ttw and test. :-/  And making sure not any previous version lurks in portal_resources, which could have been the cause of why Maurits didn't see any improvement at first.

thet:
@fredvd your fix here is the right catch! I'm preparing a PR which does actually unset the events before re-setting them, otherwise we might en up having the same event handler running multiple times.

thet:
Superseeded by this PR: https://github.com/plone/mockup/pull/1041

PR #1040, opened 2021-02-17
Fix manifest to not include node_modules, even when later removed.

mauritsvanrees:
The coredev buildout took very long for me.
Trying it manually, `python setup.py egg_info` took five minutes.
Problem was that I had a large `mockup/node_modules` directory.
This was pruned in `MANIFEST.in`, so it never ended up in the release.
But is *was* first included before being pruned, and this took extremely long.

For explanation of what happens, see the [packaging docs](https://packaging.python.org/guides/using-manifest-in/) and look for this line: "Commands are processed in the order they appear in the MANIFEST.in file."
For the mockup case it meant:

- `recursive-include mockup *.js`: include all javascript files, also when found in `mockup/node_modules`.
- `prune mockup/node_modules`: throw them away again.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,24 +1,26 @@
+# Include files that are directly in the top level directory.
 include *.json
 include *.rst
 include .yarnrc
-recursive-include mockup *.html
-recursive-include mockup *.js
-recursive-include mockup *.jshintrc
-recursive-include mockup *.json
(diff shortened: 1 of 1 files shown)


thet (approved):
LGTM

Note: When the changes from https://github.com/plone/mockup/pull/1025 make it into master we need to adapt the MANIFEST.in file again.

Issue #1016, opened 2020-09-29
Wrong context in folder_contents view


cekk:
Tried with last buildout.coredev.

Steps to reproduce:

- Create some folders structure: /Plone/aaa/bbb/ccc
- Go to portal root -> folder_contents
- navigate through sections until "ccc" folder
- Try to add a new Document

The url will be something like: http://localhost:8080/Plone/aaa/++add++Document

So the context in the sidebar is not updated with the current context. Page title, breadcrumbs and url are correctly set.

cekk:
does anyone can reproduce this bug?

mauritsvanrees:
This issue, or variations thereof, was reported several times. Also, it was claimed to have been fixed several times before, but it seems a beast that raises its head again and again, or in slightly different places.
Anyway, it *should* have been fixed by @fredvd recently, see PR #1028 and #1030, depending on which branch you use. Well, 3.x is for Plone 5.2.
Can you retry on current coredev or on Plone 5.2.3? Both contain his fixes.

fredvd:
There is still one other issue left.  Or actually two.  The drag&drop 'upload' modal also doesn't get updated, so when you navigate a fiw times, and upload something through it, it goes to the wrong folder. 

And 'where smoke there is fire', I noticed that the asterisk for marking the default item in the folder_contents is no longer shown in Plone 5.2.3, It is in Plone 5.2.2 .  But I have no idea yet if one of my fixes introduced this.   https://github.com/plone/Products.CMFPlone/issues/3220

fredvd:
Will be fixed in Plone 5.2.4.  See https://github.com/plone/mockup/pull/1039

PR #1041, opened 2021-02-19
Correctly update the context info when changing paths.

thet:
Do only remove the correct event listener on ``context-info-loaded`` before adding a new one.
Fixes a problem where the current path was not updated for the upload popup when changing paths.
Fixes: #1016
Refs: #1028, #1030, #1039

### DIFF
--- a/1041.bugfix
+++ b/1041.bugfix
@@ -1,0 +1,5 @@
+Do only remove the correct event listener on ``context-info-loaded`` before adding a new one.
+Fixes a problem where the current path was not updated for the upload popup when changing paths.
+Fixes: #1016
+Refs: #1028, #1030, #1039
+[thet]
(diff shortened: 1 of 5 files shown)


fredvd (approved):
LGTM!

In hindsight, instead of passing the function, I used the 'namespace' last year to fix other event listeners. That could have worked here as well: https://github.com/plone/mockup/blob/bef28e95c374f645c8b0bdbb4ee33780f1e80a10/mockup/patterns/structure/js/views/textfilter.js#L44

But that was done on browsers events like 'click', not our own custom events.  Passing the function is clearer instead of the namespace magic of jQuery.

thet:
The namespace thing is really magic and not quite self-explanatory.
Tnx for reviewing!

mauritsvanrees:
Released in `mockup 3.2.5` and `plone.staticresources 1.4.2`. Thank you both!

PR #1042, opened 2021-03-12
Remove fonts from patterns to avoid multiple inline includes.

agitator:
Remove fonts from patterns to avoid multiple inline includes.

Glyphicons are removed as less-includes from patterns.
In Plone we add two more bundles to include the icons: https://github.com/plone/plone.staticresources/pull/131
In Mockup docs the bootstrap.less is included, so the demo patterns still have their icons: https://github.com/plone/mockup/blob/3.x/mockup/less/docs.less#L5


thet (approved):
LGTM

mauritsvanrees (approved):
LGTM

PR #1044, opened 2021-03-16
Fix for sort_reversed checkbox in pat-querystring

### LINKED ISSUE #1043: querystring sort_reversed does not get current saved value in pat-querystring
therefore not saveable ... see https://github.com/plone/Products.CMFPlone/issues/3263

it's because the widget markup did change in Barceloneta LTS ...

petschki:
fixes #1043

### DIFF
--- a/mockup/patterns/querystring/pattern.js
+++ b/mockup/patterns/querystring/pattern.js
@@ -913,12 +913,12 @@ define([
                     self.refreshPreviewEvent.call(self);
                     if ($(this).prop("checked")) {
                         $(
-                            '.option input[type="checkbox"]',
+                            'input[type="checkbox"]',
                             existingSortOrder
                         ).prop("checked", true);
                     } else {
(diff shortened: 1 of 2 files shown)


thet (commented):
@petschki @pbauer LGTM, if it works for Plone 5.2 and Plone 6. Does it?
Do the tests pass for you? Travis obviously not, but locally? For me they don't for unrelated reasons.
Why is this change even necessary?

petschki:
This fix is necessary, because the widget markup in Plone6 Classic Theme did change ... this generalizes the determination of the checked value of `sort_reversed` for Plone5 and 6. Though I did not run the tests ... only tested in a productive project of us.

thet (approved):
ok, since this targets the master branch it only needs to work in plone 6.
looks good to me.
you're in charge of getting anything green if something breaks ;)

PR #1053, opened 2021-05-26
call fileUpload view explict with @@ to avoid rest clashes

jensens:
as in #1052


thet (approved):
LGTM

mauritsvanrees:
Does this actually have an effect?

I am trying to compile bundles for a new release for `plone.staticresources`, and this is the only change in mockup 1.x. When I try to follow the steps in the staticresources readme, the only effect is that `src/plone/staticresources/static/components/mockup/mockup/patterns/relateditems/upload.js` is changed, but as far as I know, files in the `components` directory are not used in production, except probably when you switch the resources to development mode.
Compiling the bundles only makes a change in the minified versions and map files, when you ignore the different directories (basically `/home/_thet` from when Johannes compiled them, versus `/Users/maurits` now). Well, I  see that `thememapper-compiled.css` has a change, but I have no idea what it is and it should be unrelated.

In Plone 5.2.4, uploading an image in tinymce already uses `@@fileUpload` because of the [patterns settings](https://github.com/plone/Products.CMFPlone/blob/5.2.4/Products/CMFPlone/patterns/settings.py#L157).

So I wonder if there is any sense in making new releases, as in issue #1056.

mauritsvanrees:
The question still stands for [Plone 5.2.6](https://github.com/plone/Products.CMFPlone/issues/3336).

PR #1082, opened 2021-10-26
Change the regex to extract the widgets.pot file

erral:
We are changing the regex to catch multiline msgids like this one:

See the updated widgets.pot file here:
https://github.com/collective/plone.app.locales/pull/336/files#diff-7f16dfc1665a7f292e8b762726494c0c6c918470959fedb05c41affa247644f2L441


MrTango:
@erral is this for master or es6 (Plone6)?

erral:
This is for master, for the version currently in 6.0a.

But I need to recheck the Regex, and won't be able to check in the next 2 weeks.

erral:
We have checked this and there is no need to change the regex now, as there are no new messages.

PR #1104, opened 2021-12-30
add contentrule pattern to controlpanels

MrTango:
needs https://github.com/plone/plone.app.contentrules/pull/70


MrTango:
@agitator i could use some help with the bootstrap layouting part here. Could you play a bit with the template in plone.app.contentrules so that it's looking similar to what we have now, but using bootstrap markup?
I added already some grid classes.
I would like to move on the the next pattern then.

agitator:
I'll have a look

thet (approved):
LGTM

PR #1106, opened 2022-01-12
Add discussion pattern

MrTango:


MrTango:
cleanup of https://github.com/plone/plone.app.discussion/pull/190 should be merged after.

PR #1138, opened 2022-03-29
Fix structure pattern

agitator:


thet (changes requested):
@frapell did some commits addressing the missing icons and the ``resolveIcon`` icon cache. With his commits this PR should be unnecessary. even if the resolveIcon view is called 100 or 1000 of times on a large dataset this should not really make a difference to a static icon configured directly in the view.

@agitator this PR would basically revert franco's changes.

@agitator if you think your PR makes a performance difference, please give me some numbers on how faster rendering a large table in folder_contents is with this PR compared the current mockup master (already compiled in plone.staticresources master). Or give me some instructions or a Data.fs to check it myself.

agitator:
@thet create a folder with 30+ images and you'll immediately feel it ;-)

frapell:
@thet My fix from yesterday was so icons won't be broken, however I overlooked the other issues with the folder_contents:

1. Drag and drop to change order doesn't work
2. Clearly there is some issue with async and the table rendering, since the order in which items show up is not always the same, and there's the "No data available in table" at the top.
3. Performance issues

The only thing I don't like about this PR is the SVG's being hardcoded, I would like to have something like `import("++plone++bootstrap-icons/eye.svg")` instead.

Other than that, this PR does fix the mentioned issues...

thet:
OK, let's merge then!

thet:
Obsolete via https://github.com/plone/mockup/pull/1144

PR #1139, opened 2022-03-29
Use conventional commits for automatic changelog generation

thet:
I want to use conventional commits for automatic changelog generation, as we do it in Patternslib. See:
https://github.com/Patternslib/Patterns/blob/master/CHANGES.md

Please review, if you understand the docs in the README and are all OK with this change.

We cannot use towncryer here as this is a non-Python based package. Also conventional commits has the advantage to think a bit more about the commits and have a cleaner commit history.

@mauritsvanrees
@MrTango @frapell @agitator @petschki @jensens @pbauer


petschki (approved):
OK 👍🏼

thet:
@mauritsvanrees that's right - there is something wrong with the current CHANGES file. I will check what this is...
But at least the releases page is correct - another benefit of using ``release-it`` and conventional changelog:
https://github.com/Patternslib/Patterns/releases

@mauritsvanrees I removed the news folder because I already edited the CHANGES.md manually and tried to include all relevant changes.

@frapell yes, we should construct our commits in a way that all changes which belong to a specific feature or bugfix go into only one commit. 
I make an exception for all automatically generated files. Compiled bundles go into an extra commit saying ``git add src/ && git commit -m"Compiled." -n``, yarn.lock changes go into ``git commit yarn.lock -m"yarn install." -n`` - or like so. That way I can just drop those commits when rebasing - automatically generated files tend to make always merge problems....

Since I introduced that in Patternslib I'm super happy with that process. Except when it doesn't work. But that doesn't happen very often.

@mauritsvanrees theoretically we can also adapt conventional commits in Python packages, as we could use towncryer here. Just saying :)

PR #1140, opened 2022-03-30
fix(pat validation): Customize the validation template to be Bootstra…

thet:
…p compatible.

Screenshot from 2022-03-30 11-56-21


petschki (approved):
works together with https://github.com/plone/plonetheme.barceloneta/pull/273

PR #1142, opened 2022-03-30
fix(core utils resolveIcon): Fix resolveIcon cache.

thet:
Fix the ``resolveIcon`` cache in a way that multiple asynchronous calls for the same icon does not lead to multiple network connections to the same resource.
Instead the icon cache is prefilled with the Promise which will resolve the icon and then we await the fullfilment of the Promise.

### DIFF
--- a/src/core/utils.js
+++ b/src/core/utils.js
@@ -1,6 +1,7 @@
 import $ from "jquery";
+import logging from "@patternslib/patternslib/src/core/logging";
 
-
+const logger = logging.getLogger("core utils");
 
 var QueryHelper = function (options) {
     /* if pattern argument provided, it can implement the interface of:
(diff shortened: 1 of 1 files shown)


petschki (approved):
I can confirm that icons are loading and get cached correctly which this PR is for. But there are still issues:

1. order of folder contents isn't persistent and "No data available..." information as first row
2. drag&drop reordering doesn't work every time

which are only solved when placing actionmenu icons statically ... so there needs to me more investigations on that here: https://github.com/plone/mockup/pull/1138

frapell:
@thet I can confirm this branch works faster since icons are only fetched once. Now, the issue with the async appears to be happening here
https://github.com/plone/mockup/blob/master/src/pat/structure/js/views/table.js#L84-L122
I am placing breakpoints and I see that line 122 `registry.scan(this.$el)` is called before the table gets rendered.

It is hard to wrap my head around this whole async execution, and correct me if I am wrong, but shouldn't this `registry.scan` call be called after the whole table has finished loading?

When icons were being fetched several times and loading took longer, I was seeing the bugs mentioned by @petschki every single refresh, but now with the loading being faster, I only see it occasionally

thet:
@petschki @frapell I wonder how the fix in https://github.com/plone/mockup/pull/1138 would solve these issues.
But I'm not saying we shouldn't merge the other PR too.

I'm going to merge this one.

frapell:
@petschki @thet It appears to be a matter of loading time. The "hardcoded SVG" load super fast and so it has no issues. You can place a breakpoint in https://github.com/plone/mockup/blob/689fb76d245f200ecee4f21ef92cf3e24bbe8642/src/pat/structure/js/views/table.js#L117 and take 2 seconds to continue, and you will see that after the table loads, there's the "No data available in table" at the top

PR #1144, opened 2022-03-30
Fix structure datatable

thet:


petschki (approved):
👍🏼 tested with 50 items in folder_contents -> works for me and makes #1138 obsolete

PR #1149, opened 2022-03-31
Simplify structure tablerow

thet:
This slightly reduces the complexity of the code for waiting for table row rendering end.
It basically removes one event handler.


petschki (approved):
👍 tested locally and works for me

PR #1151, opened 2022-04-01
Async rendering fixes

petschki:


src/pat/structure/js/views/app.js
```diff
@@ -626,9 +623,13 @@ export default BaseView.extend({
...
 
-        await this.tableView.render();
-        this.$el.append(this.tableView.el);
-        this.$el.append(this.pagingView.render().el);
+        const tableView = new TableView({ app: this });
+        await tableView.render()
+        this.$el.append(tableView.el);
+
+        const pagingView = new PagingView({ app: this });
+        pagingView.render()
```
thet, line 631: the pagingview is probably not async, therefore no await, right?

thet (approved):
Please squash both commits and use a commit message like "fix(pat structure): Fix async rendering." - unless you want the commit message to not be added in the changelog, which would also make sense.
for the first mockup 5.0 release we will actually maintain the changelog manually, as there are too much changes.

other than that, LGTM!

src/pat/structure/js/views/app.js
petschki: there's no async in PagingView

thet (approved):

PR #1152, opened 2022-04-01
[3.x] pat-structure: fix duplicated initialization of result collection in

petschki:
this fixes https://github.com/plone/Products.CMFPlone/issues/3464 for Plone 5.2


petschki:
see https://github.com/plone/plone.staticresources/pull/210

PR #1154, opened 2022-04-03
Update to backbone.paginator 2.x

petschki:


petschki:
work continued here #1172

PR #1157, opened 2022-04-07
feat(LiveSearch with images): Added support for images in liveSearch …

agitator:
…results and moved positioning and styling to barceloneta.


mauritsvanrees (approved):
Approved (but only tested in compiled form in `plone.staticresources`).

PR #1163, opened 2022-04-28
fix selecting columns for pat-structure datatable

petschki:
I broke the implememtation of `this.tableView` and fixed it again 🙈


MrTango:
looks like we need tests here, also the test setup still fails and has requirejs in it.
for this pattern we really need some browser testing, otherwise we lose control.

MrTango:
feel free to merge your patch, ;)

PR #1167, opened 2022-05-13
Expose patternslib registry.

maethu:
I would like to able to `re-scan` changed DOM from a 3rd party JS lib (depending on certain events). 
For thus purpose I think I have to expose the patternslib registry.

I'm not super familiar  with patternslib and mockup, so I don't know if this may expose something which should not?
I'm also happy for other suggestions on how to access the patternslib from Plone. As far I can see it's not possible right now.  

Background:
I'm working on a Plone 6 plugin and I'm currently using Plone 6.0.0a4

### DIFF
--- a/src/patterns.js
+++ b/src/patterns.js
@@ -76,3 +76,4 @@ validation_parser.parameters["error-template"].value =
     '<em class="invalid-feedback">${this.message}</em>';
 
 registry.init();
+window.__patternslib_registry = registry;


maethu:
> @maethu Not sure I understand, but you can check how it is done for instance in the modal pattern in https://github.com/plone/mockup/blob/master/src/pat/modal/modal.js you simply call `registry.scan` and you pass the element that should be scanned

Thanks @frapell for your fast reply.
I don't develop another pattern. I have my separate JS code in my addon package (packed with vite). 

If I depend on @patternslib/patternslib in my package.json  and I do `import registry from "@patternslib/patternslib/src/core/registry";` I obviously get a fresh instance of the registry, which is empty. 


With the change I propose any other 3rd party JS can do `window.__patternslib_registry.scan(body);`  and can have access to all load plone patterns.

frapell:
Oh, I see... @thet maybe you want to chime in here?

thet:
Your approach would work.
However, to really fix this problem we should have module federated bundles in place. I'm currently working at the Buschenschanksprint on finalizing that.
I plan to merge and release a new Patternslib version including module federation later today.
That includes a global Patterns registry, which most probably make your approach obsolete.
You can then just import the registry, run registry.scan and be done - even without your bundles being module-federation enabled.
See: https://github.com/Patternslib/Patterns/pull/923/commits/a87181c17c353e8ae934a3f5a84352ebf6eb9bf8
Let me know what you think about this.

maethu:
> Your approach would work. However, to really fix this problem we should have module federated bundles in place. I'm currently working at the Buschenschanksprint on finalizing that. I plan to merge and release a new Patternslib version including module federation later today. That includes a global Patterns registry, which most probably make your approach obsolete. You can then just import the registry, run registry.scan and be done - even without your bundles being module-federation enabled. See: [Patternslib/Patterns@a87181c](https://github.com/Patternslib/Patterns/commit/a87181c17c353e8ae934a3f5a84352ebf6eb9bf8) Let me know what you think about this.

Awesome that will solve my problem and it gives me event more options! Thanks you very much!!

maethu:
closed in favour of https://github.com/Patternslib/Patterns/commit/a87181c17c353e8ae934a3f5a84352ebf6eb9bf8

thet:
It's now also released in Patternslib 8.0.0

PR #1176, opened 2022-06-10
New badge for excluded items in pat-structure

petschki:
see https://github.com/collective/plone.app.locales/pull/348 for translation


thet (approved):
LGTM

PR #1177, opened 2022-06-21
Register bootstrap+jquery globally

thet:


petschki:
I've experimented with "eager sharing" option for the global modules... 

The main change here is, that there are no more "remote" modules `jquery` and `bootstrap` because they are marked as `eager` in the `mockup` shared library list. So they are available in the first chunk and can be globally initialized in `index.js` ... see https://github.com/plone/mockup/pull/1177/commits/1a6f3498a902de041333791a566f5b8d9e3312e1#diff-bfe9874d239014961b1ae4e89875a6155667db834a410aaaa2ebe3cf89820556R21 

my addons wich use `(function($) { ... })(jQuery)` and are registered after the plone bundle work now.

Somehow `window.bootstrap` is always `undefined` but that might be something else ... 

/cc @maethu are you able to test this change here?

petschki:
NOTE: you have to disable `bootstrap-js` and `jquery` bundle in the resourceregistry when using this ...

petschki:
Closed in favor of https://github.com/plone/mockup/pull/1185

maethu:
> /cc @maethu are you able to test this change here?

Awesome! I just tried it with https://github.com/plone/mockup/pull/1185 - Seems to work really well. 
I also do think a big advantage if this solution is, that I can use a different packaging tool, like vite. I still think the webpack module federation approach is valid and good way for dependency injection. But it might no be really compatible with bootstrap and other 3rd party libraries.

Issue #1175, opened 2022-06-07
bootstrap is no longer available globally with latest mockup release


maethu:
With the release 5.0.0-alpha.0 (before webpack module federation) there was a bootstrap object globally available.

With this I was for example able to control Modal instances via script
```
const customModal = new window.bootstrap.Modal(...)
```

This is no longer possible. I'm not sure why, 
For example jQuery ist still there (via https://github.com/plone/mockup/blob/master/src/index-jquery.js)

@thet You may have an idea?

maethu:
Followup

I have now a webpack config with module federation:
```
const path = require("path");
const { VueLoaderPlugin } = require("vue-loader");

const package_json = require("./package.json");
const mf_config = require("@patternslib/patternslib/webpack/webpack.mf");

module.exports = (env) => {
  let config = {
    entry: {
      "mylib.min": path.resolve(
        __dirname,
        "./path/to/js/js/src/main.js"
      ),
    },
    ...
    plugins: [new VueLoaderPlugin()],
    ...
    },
  };

  config.plugins.push(
    mf_config({
      filename: "mylib-remote.min.js",
      package_json: package_json,
      remote_entry: config.entry["mylib.min"],
    })
  );

  return config;
};
```

So far so good. 


**Problem**
window.bootstrap is still not available -> It is loaded though, since bootstrap registered it's plugins as JQuery plugins as well. Which is accessible as `window.JQuery.fn.modal`.

If I do `import { Modal } from "bootstrap";` in my code, which actually works kinda, but I have weird side effects. Seems like bootstrap gets initialised multiple times. The workaround via the JQuery plugin seems to work fine though. 

Everything else seems to work as expected!! Thanks so much to put in the work to introduce module federation!

petschki:
stumbled upon this too today ... @thet tried to fix it here, but still not working for me: #1177

maethu:
@petschki @thet 

I think something is really odd with bootstrap 5.2, or there might be integration issues with webpacks module federeation. 

If I do `await import("bootstrap")` or `import "bootstrap"` anywhere in my code. DropDowns for example do no longer work. 

The mentioned workaround via jQuery does not always work, since it does depend load order.

petschki:
I do not know how to solve this yet but it seems to me that registering modules globally and MF configured modules are conceptually a bit against each other.

[...]

PR #1185, opened 2022-07-07
Register jQuery and Bootstrap globally - alternative approach

thet:
Here I register jQuery and Bootstrap globally directly in the main bundle - making the jquery and bootstrap bundles obsolete.

jquery and bootstrap are available before DOMContentLoaded.

This increases the initial bundle size significantly:‌ from 93k to 269k.

Some thoughts:
- We could split out Bootstrap into a own bundle without module federation and provide it optionally (but we're using some bootstrap js in mockup already, so there will be some code duplication).
- We could provide two alternative main bundles - one with the global registrations and one without. Integrators can then choose which one they use.
- We don't care about the bundle size - < 300k are not that awful much in the end...
- For customized sites we can always build our own bundle, optimize it for size and not use the default Plone bundle.

If we go with this approach we need to remove the bootstrap and jquery bundles from plone.staticresources too and provide an upgrade step.

Any thoughts @petschki @MrTango @agitator ?

Fixes: https://github.com/plone/mockup/issues/1175

### DIFF
--- a/src/index-bootstrap.js
+++ b/src/index-bootstrap.js
@@ -1,2 +1,0 @@
-// Bundle entry for Bootstrap
-import "bootstrap";
--- a/src/index-jquery.js
+++ b/src/index-jquery.js
@@ -1,2 +1,0 @@
-// Bundle entry for jQuery
-import "@patternslib/patternslib/src/globals";
--- a/src/index.js
+++ b/src/index.js
@@ -20,3 +20,16 @@ import "@patternslib/patternslib/webpack/module_federation";
 
 // And now load this bundle's actual entry point.
 import("./patterns");
+
+// Register Bootstrap and jQuery gloablly
+async function register_global_libraries() {
+    // Register Bootstrap globally
+    const bootstrap = await import("bootstrap");
+    window.bootstrap = bootstrap;
+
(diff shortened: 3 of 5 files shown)


petschki (approved):
Glad to see, that you basically came to the same solution like me here https://github.com/plone/mockup/pull/1177 ... the changes here are minimal and work for me nicely with the "old" addons.

I'd prefer to deliver `jQuery` and  `bootstrap` globally OOTB since existing addons can be easier transfered to Plone 6.0 and do not break initially when they're not rewritten to module federation.

Delivering an extra bundle with module federated `jQuery` and `bootstrap` sounds like a good idea, but maybe would be enough to  document how to build such a bundle on your own if you need it.

petschki:
I prepared `plone.staticresources` with this branch and an upgrade step and started Jenkins with these changes. Once this is merged and released, I'll complete it there: https://github.com/plone/plone.staticresources/pull/231

thet:
Yes this approach is very similar. The eager keyword for the shared object was a good one!

petschki:
tests are all green https://github.com/plone/plone.staticresources/pull/231 ... I'd merge and release!

PR #1188, opened 2022-07-11
Revert async init in `pat-modal`

petschki:


thet (approved):
LGTM.

Currently the easiest way to fix this problem without adding async/awaits all over the place.

PR #1191, opened 2022-07-11
feat(pat depends): Add pat-depends from @patternslib/patternslib to the build.

thet:
This allows for optional form fields or other HTML‌ structure which depen on the value of form inputs.


petschki (approved):
👍🏼

PR #1195, opened 2022-07-15
Make `pat-relateditems`templates customizeable again via patternoptions

### LINKED ISSUE #1194: Templates for `pat-relateditems` are not customizable anymore
In an addon I've customized the `selectionTemplate` of `pat-relateditems` via patternoptions. This is a nice feature which unfortunately doesn't work anymore since this commit https://github.com/plone/mockup/commit/f7292a350db29c62714725cafc12fa4462bfc015

petschki:
fixes #1194


thet (approved):
LGTM.
sorry for removing that option in the first place.

in an upcoming optimization we could not load the default template if a customized template is available.

Issue #1201, opened 2022-07-26
Schemaeditor breaks with fieldsets


frapell:
This was reported originally in https://github.com/collective/collective.easyform/issues/345 but it is a bug in the schemaeditor pattern

PR #1203, opened 2022-07-27
Fix schemaeditor

frapell:
Fix adding fields to fieldsets, and drag and drops between fieldsets. 

Prevent adding fields to fieldsets from behaviors.

fixes gh-1201

### DIFF
--- a/src/pat/controlpanels/schemaeditor--implementation.js
+++ b/src/pat/controlpanels/schemaeditor--implementation.js
@@ -25,14 +25,20 @@ export default class SchemaEditor {
     init_add_field(fieldset_id) {
         if (fieldset_id) {
             var $add_field = $("a#add-field", this.el);
-            var href = $add_field.attr("href").split("?")[0]; // get base href without any previously set ``fieldset_id``.
-            $add_field.attr("href", href + "?fieldset_id=" + fieldset_id);
+            var $fieldset = $("#form fieldset#fieldset-"+fieldset_id, this.el);
+            if ($fieldset.data('can-add-fields') === false){
+                $add_field.hide();
+            } else {
+                var href = $add_field.attr("href").split("?")[0]; // get base href without any previously set ``fieldset_id``.
+                $add_field.attr("href", href + "?fieldset_id=" + fieldset_id);
+                $add_field.show();
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


frapell:
@petschki @thet You guys want to give it a try?

Btw, I am preventing adding and drag&drop fields to fieldsets added from behaviors, since I don't think that was ever intended... At least I am not finding a proper way to do it...

You will need to use the mentioned branch from https://github.com/plone/plone.schemaeditor/pull/92

src/pat/controlpanels/schemaeditor--implementation.js
```diff
@@ -25,14 +25,20 @@ export default class SchemaEditor {
     init_add_field(fieldset_id) {
         if (fieldset_id) {
             var $add_field = $("a#add-field", this.el);
-            var href = $add_field.attr("href").split("?")[0]; // get base href without any previously set ``fieldset_id``.
-            $add_field.attr("href", href + "?fieldset_id=" + fieldset_id);
+            var $fieldset = $("#form fieldset#fieldset-"+fieldset_id, this.el);
+            if ($fieldset.data('can-add-fields') === false){
+                $add_field.hide();
```
petschki, line 30: I've tested this together with https://github.com/plone/plone.schemaeditor/pull/92 and it works like a charm. I have one suggestion: I would use the css class `disabled` for the `Add field` button when its prevented from adding instead of removing it. This also prevents clicking the button.

petschki (changes requested):

src/pat/controlpanels/schemaeditor--implementation.js
frapell: @petschki Much better, indeed :)

frapell:
@thet Do I need to do anything to the changelog? I was under the impression that it was automatically generated out of the commits when they include parts like `fix(pat schemaeditor)` right?

petschki:
Nothing to do in changelog since your commit message fits the release-it conventional changelog ...

I suggest to do the following:
Create a PR in `plone.staticresources` with the latest commit hash here as mockup dependency  in `package.json` and Test it together with the schemaeditor PR in Jenkins.

petschki (approved):
LGTM

PR #1211, opened 2022-08-26
fix(pat-tinymce): set the language in tiny options

### LINKED ISSUE #1209: TinyMCE translations are not loaded in latest Plone 6b1
In Plone 6b1 as it is in https://6-classic.demo.plone.org/, TinyMCE translations are not correctly loaded. 

See this screenshot of the spanish part of the site: all menus and menu items are in English.

Pantaila-argazkia 2022-08-12 20-04-52

The same happens in German:

Pantaila-argazkia 2022-08-12 20-05-43

Nevertheless I see that the `tinymce-i18n` package [is added as a dependency in mockup](https://github.com/plone/mockup/blob/master/package.json#L41-L42), so there might be a problem loading those translations.

erral:
After some debugging I have noticed that although the correct language js was being loaded, we need to explicitly set the language we are using in tiny options.

This way the correct language js is applied and the editing interface shows translated. See:

German
Pantaila-argazkia 2022-08-26 14-59-51

Spanish 
Pantaila-argazkia 2022-08-26 15-00-44

Basque
Pantaila-argazkia 2022-08-26 15-00-22

Fixes #1209 and https://github.com/plone/Products.CMFPlone/issues/3619

### DIFF
--- a/src/pat/tinymce/tinymce.js
+++ b/src/pat/tinymce/tinymce.js
@@ -1,4 +1,5 @@
 import Base from "@patternslib/patternslib/src/core/base";
+import I18n from "../../core/i18n";
 import _t from "../../core/i18n-wrapper";
 
 export default Base.extend({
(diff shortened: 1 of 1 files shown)


davisagli:
Needs changelog.

Hmm, there is also a failing test, but it doesn't look related

erral:
the readme says that the changelog is handled automatically 🙄 that's why I didn't add anything

davisagli:
@erral oh hmm! It's certainly possible that this repo handles it differently and that I'm not aware of it -- but I'm not sure, since there is also a changelog verifier action

MrTango:
indeed the changelog is generated by the commit message

MrTango:
@thet when are the commit messages are converted into the changelog? Is there something missing hre or do we just need to update mr.roboto checks?

MrTango (approved):
LGTM

thet:
@MrTango the changelog is generated via release-it and conventional-changelog when a release is done.
When installing mockup via ``make install`` the a pre-commit hook would be installed via ``husky`` which would check the formatting of the commit message via ``commitlint``.

How can we remove the changelog verifier for pull requests?

PR #1212, opened 2022-08-26
fix(pat-select2): load language files

### LINKED ISSUE #994: Select2 based widgets show always english messages
Both in folder_contents and in the edit form. 

See:

2020-07-03 09-57-55-ko-pantaila-argazkia

2020-07-03 09-58-13-ko-pantaila-argazkia

erral:
Fixes #994. Similar to #1211 we need to explicitly set the language and load the language js file

Basque:
Pantaila-argazkia 2022-08-26 15-30-02

German
Pantaila-argazkia 2022-08-26 15-28-53

Spanish:
Pantaila-argazkia 2022-08-26 15-29-42

### DIFF
--- a/src/pat/select2/select2.js
+++ b/src/pat/select2/select2.js
@@ -1,5 +1,6 @@
 import $ from "jquery";
 import Base from "@patternslib/patternslib/src/core/base";
+import I18n from "../../core/i18n";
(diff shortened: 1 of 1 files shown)


MrTango (approved):
LGTM

Issue #1205, opened 2022-08-04
relateditems: not orderable on existing content


mauritsvanrees:
Tried in current Plone coredev 6, both with and without mockup in development mode, and in Plone 6.0.0a3 with the standard plone.staticresources for that version.

- Create a new page, do not save it yet.
- In the related items field add News and Events.
- Change the order of the two. This works.
- Save the page.
- Edit the page.
- Try to change the order of the two related items again. This fails. Nothing happens. The click is interpreted as: you want to add a new related item.

petschki:
I can confirm on 6-classic.demo.plone.org ... but when editing the page, once you've added a new item, you can resort them again. Seems there are some initial select2 events missing when loading the page.

PR #1214, opened 2022-08-30
fix(pat-select2): initialize sorting, fixes issue 1205.

mauritsvanrees:
This fixes https://github.com/plone/mockup/issues/1205, that related items are not sortable on edit.

This is probably not the cleanest fix, but it is only one line and gets the job done, so perhaps it is okay. If someone knows a cleaner way, I am happy to hear it.

Let me try to explain why this works:

- At the [end of the init function](https://github.com/plone/mockup/blob/5.0.0-alpha.21/src/pat/select2/select2.js#L240-L241) we first call `self.initializeOrdering()` and then `await self.initializeSelect2()`.
- `initializeOrdering` sets up an [on change event handler](https://github.com/plone/mockup/blob/000884f123b3f0e09040960010be07ef8d21f657/src/pat/select2/select2.js#L87) that uses `select2`.
- So at the end of `initializeSelect2`, I trigger a change event on the main element, and this calls the event handler, making sure the selected items can be ordered.

Cleaner would probably be to move the code from the event handler into a separate method, and call this function on change and at the end of `initializeSelect2`. But when I tried this, javascript complained that my function did not exist. Probably some misunderstanding on my end.

### DIFF
--- a/src/pat/select2/select2.js
+++ b/src/pat/select2/select2.js
@@ -150,6 +150,7 @@ export default Base.extend({
         self.$el.parent().off("close.plone-modal.patterns");
         if (self.options.orderable) {
             self.$select2.addClass("select2-orderable");
+            self.$el.change();
         }
     },
     opened: function () {


thet (changes requested):
@mauritsvanrees please consider this PR which is based on your analysis: https://github.com/plone/mockup/pull/1216

mauritsvanrees:
Closed in favour of PR #1216.

PR #1216, opened 2022-09-01
Fix select2/relateditems sorting issue

thet:
Alternative approach to: https://github.com/plone/mockup/pull/1214
Fixes: https://github.com/plone/mockup/issues/1205

@mauritsvanrees based on your analysis I built an alternative, more appropriate approach to the sorting issue.
``initializeOrdering`` is called after ``initializeSelect2`` and also initializes existing items without a ``change`` event.
I added you as co-author.

@mauritsvanrees Please review, test and eventually approve.


mauritsvanrees (approved):
I can more or less follow the code, and it seems okay.
Testing it now: yes, it works. Thanks!

PR #1217, opened 2022-09-08
Rework navigationmarker

thet:
- Modernized code
- Simpler, more efficient and flexible navigationmarker
- Configurable in-path and current classes
- Some other changes along the way...


petschki:
@thet I think we can merge this safely ?

PR #1218, opened 2022-09-09
feat: pat-tinymce: add table header by default

MrTango:


MrTango:
right now it only works when you add a extra button 'inserttable' to tinymce, but it should also work for the 'inserttable' in the table menu. I ask here, let's see what we can do there. https://github.com/tinymce/tinymce/issues/6472

src/pat/tinymce/tinymce--implementation.js
```diff
@@ -219,6 +219,24 @@ export default class TinyMCE {
                 self.tiny = editor;
             }
         };
+        tinyOptions["setup"] = (editor) => {
+            editor.ui.registry.addMenuButton('inserttable', {
+                icon: 'table',
+                tooltip: 'Insert table 2',
```
thet, line 225: Can we have a better title, maybe "Insert table with table headers" ?
We should also add the option to translate this with the i18n helpers.

petschki (approved):

thet (commented):
Looking at the code it seems to me that this button is added automatically.
If not, how can we add it?

PR #1227, opened 2022-10-15
[testfix] pat-markspeciallinks

petschki:


src/pat/markspeciallinks/markspeciallinks.js
```diff
@@ -94,7 +93,7 @@ export default Base.extend({
                 if (icon_name) {
                     const icon = await utils.resolveIcon(icon_name);
                     const icon_el = dom.create_from_string(icon);
-                    icon_el.classList.add("markspeciallinks__icon");
+                    icon_el.querySelector("svg").classList.add("markspeciallinks__icon");
```
petschki, line 96: @thet question: `dom.create_from_strinng` returns a `DocumentFragment` which doesn't have `classList` in my jest tests ... is this solution correct or is there anything else I could do?

thet: that looks correct

thet (commented):
:tada:

PR #1234, opened 2022-10-28
Fix editing weekly weekdays

petschki:
Editing weekly weekdays was broken due to array index/value mismatch in input name.

This also fixes reloading occurrence when changeing the weekdays checkboxes.

### DIFF
--- a/src/pat/recurrence/recurrence.js
+++ b/src/pat/recurrence/recurrence.js
@@ -76,9 +76,7 @@ function widgetSaveToRfc5545(form, conf, tz) {
                 weekdays = "";
                 i18nweekdays = "";
                 for (j = 0; j < conf.weekdays.length; j++) {
-                    input = field.find(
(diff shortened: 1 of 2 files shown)


src/pat/recurrence/recurrence.js
```diff
@@ -660,7 +660,7 @@ const RecurrenceInput = function (conf, textarea) {
     var now_date = new Date().toISOString().substring(0, 10);
     var index, i;
 
-    for (i = 0; i < 7; i++) {
+    for (let i = 0; i < 7; i++) {
```
thet, line 663: the initialization of `var i` above should be removed.

thet (changes requested):
LGTM except for the minor change mentioned in the review comment.

src/pat/recurrence/recurrence.js
petschki: I've removed every obsolete declarations.

PR #1237, opened 2022-11-10
Fix webpack cachegroups

thet:


petschki (approved):
tested this branch with locally compiled staticresources and works perfect!

PR #1238, opened 2022-11-16
fix: remove picture-variants defaults, they mess up the order

MrTango:
when combined with new variants from the registry.


petschki (approved):
👍🏼

Issue #1254, opened 2022-12-08
'make install' fails


mauritsvanrees:
TLDR: we need to define `YARN` in our Makefile. I will make a PR.

```
$ make install
install
usage: install [-bCcpSsv] [-B suffix] [-f flags] [-g group] [-m mode]
               [-o owner] file1 file2
       install [-bCcpSsv] [-B suffix] [-f flags] [-g group] [-m mode]
               [-o owner] file1 ... fileN directory
       install -d [-v] [-g group] [-m mode] [-o owner] directory ...
make: *** [install] Error 64
```

Looks like this is because `make install` calls `$(YARN) install` and `YARN` is not defined. Do other developers have this defined in `.env`, which is included in the Makefile?

When I edit `Makefile` and add `YARN=yarn` it works.

```
$ make install
yarn install
yarn install v1.22.5
...
touch stamp-yarn
```

Hey, when I remove my edit and run again, it unexpectedly keeps working, but with extra warnings:

```
$ make install
Makefile:23: warning: overriding commands for target `stamp-yarn'
node_modules/@patternslib/dev/Makefile:23: warning: ignoring old commands for target `stamp-yarn'
Makefile:23: warning: overriding commands for target `install'
node_modules/@patternslib/dev/Makefile:23: warning: ignoring old commands for target `install'
Makefile:34: warning: overriding commands for target `eslint'
node_modules/@patternslib/dev/Makefile:41: warning: ignoring old commands for target `eslint'
Makefile:36: warning: overriding commands for target `check'
node_modules/@patternslib/dev/Makefile:46: warning: ignoring old commands for target `check'
Makefile:51: warning: overriding commands for target `bundle'
node_modules/@patternslib/dev/Makefile:61: warning: ignoring old commands for target `bundle'
Makefile:56: warning: overriding commands for target `bundle-pre'
node_modules/@patternslib/dev/Makefile:50: warning: ignoring old commands for target `bundle-pre'
npx yarn install
yarn install v1.22.19
...
touch stamp-yarn
```

Aha, this works because our Make file does: `-include node_modules/@patternslib/dev/Makefile`. And this file defines `YARN ?= npx yarn`.

So we should define `YARN` ourselves, otherwise the first `make install` will fail. I will make a PR.

petschki:
fixed in #1255

PR #1255, opened 2022-12-08
maint: define YARN in our Makefile.

mauritsvanrees:
Fixes issue #1254.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,7 @@
 -include .env
 export
 
+YARN   ?= npx yarn
 
 .PHONY: install
 stamp-yarn install:


petschki:
Is this really needed? `YARN ?= npx yarn` is already defined in the base Makefile here: https://github.com/Patternslib/dev/blob/main/Makefile#L15

petschki:
ah ... sorry, saw the issue too late ... never had this. maybe because of `nvm` installed here...

mauritsvanrees:
I think you will have the same issue in a fresh clone, or when you remove node_modules. I can't imagine how it works without this.
I use nvm as well.

petschki:
confirmed. fresh clone raises error. this PR fixes it. I'll merge.

PR #1260, opened 2022-12-09
Fix close selector for default cancel buttons

petschki:
Nice case-insensitive CSS selector: https://caniuse.com/css-case-insensitive ... though works only with ES6 `querySelectorAll` and not jQuery.


mauritsvanrees (commented):
This works  when viewing the full registry list or after going to another page with batching, but not when you first use the filter. Then only the second click works.

petschki:
Hm ... workes for me in every case ... can u try again? I've rebased this on the latest changes on master.

Issue #1253, opened 2022-12-07
`pat-tinymce`: Provide a proper fallback when a language is not found


frapell:
This is coming from https://github.com/plone/plone.app.mosaic/issues/511

When the site is set to a language for which there is no specific translation (Example with en_US), then tinymce is not properly initialized. 
I haven't found yet why this seems to be a problem when there are multiple instances of tinymce in "inline mode" (as described in the original ticket in mosaic), and it doesn't seem to cause any apparent issues for when tinymce is not used in "inline mode" (Like the default case for the "text" field in a normal add/edit form), other than a 404 when trying to get the language .JS file.

In any case, it should be properly fixed, as a suggestion I believe it first should try to find the proper language with the variant, if not found, try the language without variant, and finally fallback to "en".

PR #1261, opened 2022-12-09
`pat tinymce` Improve tinymce async initialization

frapell:
This PR fixes gh-1253 and also I have refactored the `initLanguage` method to not use a callback, but instead properly await it.
In addition, I added a check so only valid plugins (plugins that can be imported) are included.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,8 +38,8 @@
         "js-cookie": "^3.0.1",
         "select2": "git+https://github.com/ivaynberg/select2.git#3.5.4",
         "sortablejs": "^1.15.0",
-        "tinymce": "^5.10.2",
-        "tinymce-i18n": "^22.10.18",
+        "tinymce": "^5.10.7",
+        "tinymce-i18n": "^22.12.4",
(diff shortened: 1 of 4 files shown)


frapell:
@petschki @thet If you guys could review this, I think it is important we can include these fixes before Plone 6.0 final is released...

petschki (approved):
works perfect ... nit picking: I'd change the `console.log` to `logger.debug` but otherwise approved 👍🏼

petschki:
can we pin tinymce `^5.10.7` here too? see #1256

frapell:
@petschki Yeah, I saw that pull request but wasn't sure how it is handled, should I add that pin in this PR? or should we merge that one?

petschki:
if you add the pin here, dependabot closes his PR automatically ...

frapell:
@petschki bumped the versions and replaced `console.log` with `log.debug` I squashed all commits in one and force pushed, since I messed up by using " ` " in the commit message

Issue #1198, opened 2022-07-20
Dead code in tinymce implementation


petschki:
Since mockup 5.x there's no `plone-modal-dialog` class anymore and TinyMCE 5.x removed the option `ui_container`.

https://github.com/plone/mockup/blob/f332c874bacf071834c080eab964903b910a6dad/src/pat/tinymce/tinymce--implementation.js#L281-L295

Besides that we should make sure that `pat-tinymce` inside `pat-plone-modal` works as expected in the new implementation.

petschki:
@frapell since you've touched `pat-tinymce` yesterday, could you maybe take a look at this? 🫣

PR #1262, opened 2022-12-12
maint(pat tinymce): Cleanup old code no longer needed. closes gh-1198

frapell:
@petschki I have verified, by creating a Mosaic tile with 2 RichText fields, with an edit view, instead of inline rendering.
I can confirm that `self.$el.parents(".plone-modal-dialog");` is always length 0, so that code is no longer used.

Both tinymce work fine

I created this test product, and included it in buildout.coredev along with mosaic to test https://github.com/frapell/mosaic.tiles


petschki (approved):
👍🏼

Issue #1210, opened 2022-08-24
recurrence widget: use rrule lib instead backend


thet:
We're using a endpoint on the backend to calculate the dates for a given recurrence rule.
Instead we could use this lib: https://github.com/jakubroztocil/rrule

Issue taken over from: https://github.com/collective/jquery.recurrenceinput.js/issues/22#issuecomment-31445354

PR #1268, opened 2023-01-11
Fix recurrencewidget to initialize saved value correctly and fix RDATE and EXDATE editing

### LINKED ISSUE #1266: `pat-recurrence`: selected dates don't change when "Repeats" field changes
When you add a recurrence to an event, and then change the repetition, the selected dates don't change.
You need to click in / change another field to make the selected dates list change.

Tested on Plone 6.0.0 final.

See this video, taken on demo :

recurrence

cc @petschki

petschki:
fixes #1266 

This is also partly a code cleanup of the old jquery-recurrencewidget approach. The main fix is to save the parsed `RRULE` object to the `textarea['ical']` attribute instead of the jquery form ... this gives us a sane state during editing the recurrence widget multiple times.

### DIFF
--- a/src/pat/recurrence/recurrence.js
+++ b/src/pat/recurrence/recurrence.js
@@ -44,12 +44,12 @@ function format(date, fmt, conf) {
     return result;
 }
 
-function widgetSaveToRfc5545(form, conf, tz) {
+function widgetSaveToRfc5545(form, icaldata, conf, tz) {
     var value = form.find("select[name=rirtemplate]").val();
     var rtemplate = conf.rtemplate[value];
-    var result = rtemplate.rrule;
+    var result = "RRULE:" + rtemplate.rrule;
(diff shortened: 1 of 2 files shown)


thet (approved):
:+1:

laulaz (approved):
Thanks @petschki 👍

petschki:
@laulaz  I just realized, that this not really fixes your issue when changeing the select ... give me a few more minutes ...

petschki:
now fixed ... I'll merge and release. @thet can we release this in staticresources then too? I need this for my projects too ...

petschki:
oh my ... just found out that when creating an event the ajax call to `@@json_recurrence` breaks ... I'll do a new PR with the fix. We really should use a JS library to generate the events from the RRULE instead of calling the backend ... see https://github.com/plone/mockup/issues/1210

PR #1276, opened 2023-02-07
Fix pat-sortable tests

thet:


thet:
It's already merged. GitHub had an hickup.

PR #1285, opened 2023-02-13
Correct starttime of RDATE in recurrence.

petschki:


mauritsvanrees:
Would this be a fix for https://github.com/plone/plone.app.event/issues/365 and/or https://github.com/plone/plone.app.event/issues/366 ?  Or is it unrelated?
(I have not checked those bugs, I just happened to notice them this morning.)

petschki:
This is unrelated. During my recurrence widget refactoring to ES6 I removed the RDATE time fix and re-added it here... and this is Plone 6 only. ics_view seems to be broken anyways for recurring dates.

PR #1290, opened 2023-02-14
fix: downgrading datatable due to incompability

toalba:
With the newer version of the data table there appears to be a error that passes the test. Research needed.


toalba:
`Uncaught (in promise) TypeError: n.ext is undefined
    45700 dataTables.bootstrap5.mjs:30
    Webpack 56
dataTables.bootstrap5.mjs:30
    d base.js:65
    AsyncFunctionThrow self-hosted:637
    (Async: async)
    n base.js:105
    c base.js:40
    initPattern registry.js:119
    scan registry.js:200
    render table.js:158
    InterpretGeneratorResume self-hosted:1429
    AsyncFunctionNext self-hosted:632
    Backbone 5
    jQuery 6
    Backbone 6
    initialize app.js:97
    Backbone 3
    init structure.js:160
    InterpretGeneratorResume self-hosted:1429
    AsyncFunctionNext self-hosted:632
    (Async: async)
    d base.js:59
    n base.js:105
    c base.js:40
    initPattern registry.js:119
    scan registry.js:200
    init registry.js:71
    jQuery 9
    init registry.js:64
    68212 patterns.js:73
    Webpack 6
`

JavaScript Error when running in browser

rnixx:
@thet @petschki Datatables update broke the bundle.

thet:
@toalba @rnixx thanks for the investigation and this fix!

With the information you provided here I just had to see it with my own eyes. This led to the fix in https://github.com/plone/mockup/pull/1291
I manually checked all the datatables dependencies which we have in Mockup and found that all the datatables addons claim also to be compatible with datatables 1.12.1.
Therefore I did require on datatables `<1.13` in: https://github.com/plone/mockup/pull/1291/commits/585ea8a972cba364f52c5fcf7d3927e01b9d87a4

This is basically the same fix as yours and I added you as co-authors, if you don't mind.

thet:
Closed in favor of: https://github.com/plone/mockup/pull/1291

thet:
Reopened because in plone.staticresources the resolutionDependencies also would need to be updated and that could lead to other problems in other packages depending on mockup too.
Better be safe than sorry, so we do an explicit downgrade of all the datatable packages.

PR #1292, opened 2023-03-09
fix(pat structure): do not add charset to x-www-form-urlencoded header [5.0.x]

mauritsvanrees:
Adding a charset to this Content-Type is illegal according to the definition, and fails with Zope master.

See https://github.com/plone/buildout.coredev/pull/844#issuecomment-1461620409
In that PR, several things in the folder contents are broken.  I talk about reordering that fails, but that is still broken with this mockup PR.  I don't know where the `x-www-form-urlencoded` is set for that part.

What the current PR *does* fix, is the cut, copy, etcetera actions in the folder contents table, at the end of an item.
The similar actions *above* the table already work without any changes, they do not have this problem.

### DIFF
--- a/src/pat/structure/js/actions.js
+++ b/src/pat/structure/js/actions.js
@@ -68,7 +68,7 @@ export default Backbone.Model.extend({
         const url = new URL(this.app.buttons.get(buttonName).options.url);
         const resp = await fetch(url, {
             headers: {
-                "Content-Type": "application/x-www-form-urlencoded; charset=UTF-8",
+                "Content-Type": "application/x-www-form-urlencoded",
             },
             method: "POST",
             body: new URLSearchParams({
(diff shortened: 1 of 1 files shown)


mauritsvanrees:
This would need to be ported to master (seems easy).
We could port it to the mockup used on Plone 5.2, but the Zope version there is not affected, so the current code will work.  The fixed code should work as well though.

petschki:
This is due to `$.ajax` default `contentType: "application/x-www-form-urlencoded; charset=UTF-8"` (see http://api.jquery.com/jquery.ajax/) ... this is called on reordering: https://github.com/plone/mockup/blob/5.0.x/src/pat/structure/js/views/app.js#L488

petschki:
@mauritsvanrees is this still needed? since Zope master supports now `charset` for `application/x-www-form-urlencoded` I'd leave it here too.

mauritsvanrees:
Removing it would technically be cleaner.  But the old way works and will continue working, so let's close it indeed.

PR #1293, opened 2023-03-15
Add `pat-checklist`

petschki:
With this we can re-enable `toggle-all` functionality in controlpanel templates (for example group membership panel) see https://github.com/plone/Products.CMFPlone/issues/3740


petschki:
as mentioned here https://github.com/plone/Products.CMFPlone/issues/3740 we could use this for some controlpanel management templates

thet (approved):
:+1:

PR #1304, opened 2023-04-12
fix: Do not add picture-variants classes and attributes to external i…

### LINKED ISSUE #1302: TinyMCE external images contain picture-variant attributes, which resolts in broken images
When TinyMCe inserted picture-variant attributes, the outputfilter will convert it into srcsets which is not working.
We should not add those attributes in TinyMCE.

MrTango:
…mages

fixes #1302


thet (approved):
Makes sense and LGTM

PR #1305, opened 2023-04-12
Upgrade bootstrap to 5.3.0-alpha3

frapell:


petschki:
Completely unrelated but I could tackle this: since `nwsapi=2.2.3` some selectors doesn't work anymore ... I've pinned it to `2.2.2` -> see issue here: https://github.com/dperini/nwsapi/issues/83

@thet in case you also get this in `patternslib`

gforcada:
Out of curiosity, is there any ETA for a final release of bootstrap 5.3? 🤔 the `alpha` on its name does not sound to stable 😆 specially if we are thinking of making it one of the main features of Plone 6.1 on the Classic UI part 😅

PR #1309, opened 2023-04-21
Async initialization of textareas

petschki:
This should fix TinyMCE in modals not showing up (see https://github.com/plone/plone.app.mosaic/issues/528)


thet (approved):
LGTM.

I‌wonder if this really fixes the problem of not initialized textareas, but anyways this is just the correct way. This would make sure that the pattern initialization events are thrown correctly - not before but after the pattern has been initialized.

petschki:
Not sure either. But here in my browser I could not reproduce the false initialization anymore with this fix. Without it the `text` field was randomly missing in mosaics property overlay.

PR #1314, opened 2023-04-26
Fix `pat-recurrence` default selected "range option" [master]

### LINKED ISSUE #1308: Recurrence widget does not set end recurrence by default
In Plone 6.0 the recurrence widget by default it is initialized without checking any recurrence end.

I have being comparing the old `jquery.recurrenceinput.js` with the new `recurrence.js`. But no idea where is the error.

Seems this line is not respected: https://github.com/plone/mockup/blob/master/src/pat/recurrence/templates/form.xml#L244

petschki:
fixes #1308


frapell:
LGTM!

PR #1317, opened 2023-05-03
Do not remove current paragraph when inserting image in TinyMCE [master]

petschki:
fixes https://github.com/plone/Products.CMFPlone/issues/3768


MrTango (approved):
LGTM

PR #1320, opened 2023-05-16
Fix TinyMCE external url handling

### LINKED ISSUE #1116: TinyMCE tries to be too smart with external links
When we enter an external link pointing to the current portal, TinyMCE tries to bee too smart and tries to build an internal link.

Our use-case is the following:

In our portal we have many URLs coming from an external database and their urls are traversal based URLs, for instance: `http://portal.com/en/course/slug` where `course` is the view that handles the traversal and `slug` the identifier we use to query the database. 

If we want to link that URL from some other content in Plone, we need to add the link as an external link, but then TinyMCE starts doing a lot of fancy stuff with those external links and converts them to `../../../en/course/slug` which in the end renders invalid links.

I think that when entering external links TinyMCE should just accept anything incoming and let it as is and not try to be too smart.

petschki:
Fixes #1116 

Add `urlconverter_callback` and skip converting external urls.

IMPLEMENTATION NOTES:
This is a quick and rough solution which checks if inserted url starts with "http" ... it would be more solid if we query the inserted element and check for the `dataset.linktype` value, but right now I have no clue, how to get the element in the callback.

### DIFF
--- a/src/pat/tinymce/tinymce--implementation.js
+++ b/src/pat/tinymce/tinymce--implementation.js
@@ -278,6 +278,20 @@ export default class TinyMCE {
             }
         }
 
+        // add "urlconverter_callback" to leave external URLs/Images as is
+        tinyOptions["urlconverter_callback"] = (url, node, on_save, name) => {
+            if (url.indexOf("http") === 0) {
+                // if url starts with "http" return it as is
+                return url;
+            }
+            // otherwise default tiny behavior
+            if (self.tiny.settings.relative_urls) {
+                return self.tiny.documentBaseURI.toRelative(url);
+            }
+            url = self.tiny.documentBaseURI.toAbsolute(url, self.tiny.settings.remove_script_host);
+            return url;
+        }
+
         tinymce.init(tinyOptions);
         self.tiny = tinymce.get(self.tinyId);
 


MrTango (approved):
LGTM

thet (changes requested):
I‌ think this needs some more discussion: https://github.com/plone/mockup/issues/1116#issuecomment-1549419289

src/pat/tinymce/tinymce--implementation.js
```diff
@@ -278,6 +278,20 @@ export default class TinyMCE {
...
+        tinyOptions["urlconverter_callback"] = (url, node, on_save, name) => {
+            if (url.indexOf("http") === 0) {
+                // if url starts with "http" return it as is
+                return url;
+            }
+            // otherwise default tiny behavior
+            if (self.tiny.settings.relative_urls) {
+                return self.tiny.documentBaseURI.toRelative(url);
+            }
+            url = self.tiny.documentBaseURI.toAbsolute(url, self.tiny.settings.remove_script_host);
```
thet, line 291: According to the discussion on https://github.com/plone/mockup/issues/1116#issuecomment-1549419289 ff, this can be simplified to:
```suggestion
        tinyOptions["urlconverter_callback"] = (url) => {
```

thet (approved):
LGTM, but consider to apply my suggestion.

src/pat/tinymce/tinymce--implementation.js
petschki: I've just tried that locally and this breaks `plone.outputfilters` regex matching `resolveuid` in an url. So no transformation is done for `/Plone/resolveuid/<uid>` which is returned by the plonelink/ploneimage plugin for internal links/images.

petschki: yap ... unfortunately this cannot work without further changes to `plone.outputfilters` though it's a simple non-breaking change to the `resolveuid_re` from
```
resolveuid_re = re.compile("^[./]*resolve[Uu]id/([^/]*)/?(.*)$")
```
to
```
resolveuid_re = re.compile("^.*/resolve[Uu]id/([^/]*)/?(.*)$")
```
otherwise url patterns like `/<plonesiteid>/resolveuid/<uid>` will not get resolved.

Note that this is what you get if you edit your Plone site without virtual hosting. My local virtualhosting test resolves to `/resolveuid/<uid>` which the outputfilter would match though ...

petschki: Let's be not too intrusive for now and keep this as a fix for the external urls. We can fix/generalize internal url transformation later...

thet: Actually, that kind of url converting is probably really needed. Otherwise the portal path could end up being added to the text, which is not what we want.

I was wondering why this resolveuid is [inserted](https://github.com/plone/mockup/blob/8ee638bf1910cbb880fc2f16ca22a0d2ca2953af/src/pat/tinymce/js/links.js#L123) [with the](https://github.com/plone/mockup/blob/8ee638bf1910cbb880fc2f16ca22a0d2ca2953af/src/pat/tinymce/tinymce--implementation.js#L66) [full path](https://github.com/plone/Products.CMFPlone/blob/4eba7476bfdecb15e7e0d95082f52d609d15987d/Products/CMFPlone/patterns/settings.py#L164), where in this case a relative-to-the-context URL could be inserted in the first place and thus not needing to rely on the TinyMCE `urlconverter_callback`.

I did some software archeology. The `appendToUrl` setting was [initially relative-to-the-context](https://github.com/plone/Products.CMFPlone/commit/fe237918574bea82adb62f70b0fe04749e150a1c#diff-f1df2418615671723080af2f000eb99c64e05b864da04242445f49c578efc1afR73). It become an relative URL with the full path due to the wrong assumptions in this commit: https://github.com/plone/Products.CMFPlone/commit/243773d8d1b0f1804cf6db6a92b961fa85598f84 - also see the comments there.

So, there would be another angle to fix the issue:‌ Reverting the change where the full path was inserted in CMFPlone, adding `@@resolveuid` instead only `resuloveuid` as mentioned in the comment of the commit referenced above and allowing an `@@` in the [resoloveuid_and_caption](https://github.com/plone/plone.outputfilters/blob/f2823872e257765ca7379a6b4295f49f837047c2/plone/outputfilters/filters/resolveuid_and_caption.py#L31) filter in plone.outputfilters. :exploding_head: 

Since that's getting absurdly complex, let's stick with your PR as-is!

yurj: > yap ... unfortunately this cannot work without further changes to `plone.outputfilters` though it's a simple non-breaking change to the `resolveuid_re` from
> 
> ```
> resolveuid_re = re.compile("^[./]*resolve[Uu]id/([^/]*)/?(.*)$")
> ```
> 
> to
> 
> ```
> resolveuid_re = re.compile("^.*/resolve[Uu]id/([^/]*)/?(.*)$")
> ```
> 
> otherwise url patterns like `/<plonesiteid>/resolveuid/<uid>` will not get resolved.
> 
> Note that this is what you get if you edit your Plone site without virtual hosting. My local virtualhosting test resolves to `/resolveuid/<uid>` which the outputfilter would match though ...

As in https://github.com/plone/plone.app.contenttypes/pull/675 the problem is that `plone.outputfilters` and the WYSIWYG editor (TinyMCE) does not share the same code for UID resolving. The `plone.outputfilters` uid resolver should be an util imported and exposed by a `plone.api` call which than TinyMCE can call via a browser view. The same code should be imported in `plone.restapi`. So one place, one policy, one service and everybody is happy.

PR #1332, opened 2023-07-21
prevent duplicate initialization of the dropzone.

jensens:
this fixes https://github.com/plone/Products.CMFPlone/issues/3814

### DIFF
--- a/src/pat/upload/upload.js
+++ b/src/pat/upload/upload.js
@@ -65,7 +65,7 @@ export default Base.extend({
         var self = this;
 
         if (self.$el.find(".upload-area").length) {
-            logger.info("Found already a dropzone on element, skipping!");
+            log.info("Found already a dropzone on element, skipping!");
             return;
         }
 
--- a/src/pat/upload/upload.js
+++ b/src/pat/upload/upload.js
@@ -61,6 +61,11 @@ export default Base.extend({
 
         var self = this;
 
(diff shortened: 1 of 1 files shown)


src/pat/upload/upload.js
```diff
@@ -61,6 +61,11 @@ export default Base.extend({
 
         var self = this;
 
+        if (self.$el.find(".upload-area").length) {
+            console.warn("Found already a dropzone on element, skipping!");
```
thet, line 65: I'd use the info level, otherwise people might get scared of a warning which is harmless.
Also, please use the logging infrastructure. I'd change it to:

```
import logger from "@patternslib/patternslib/src/core/logging";

const log = logger.getLogger("pat-upload");

[...]

        if (self.$el.find(".upload-area").length) {
            logger.info("Found already a dropzone on element, skipping!");
```

thet (approved):
all in all LGTM, except two suggestions:
1) see my comment in the code
2)‌ Please change the commit message to:

```
fix(pat-upload): Prevent duplicate initialization of the dropzone.

Fixes: https://github.com/plone/Products.CMFPlone/issues/3814
```

This would add this change automatically with the next release.

src/pat/upload/upload.js
jensens: thanks, done

PR #1334, opened 2023-07-28
Upgrade

thet:


package.json
```diff
@@ -34,7 +34,7 @@
         "datatables.net-select-bs5": "1.4.0",
         "dropzone": "4.3.0",
         "jqtree": "1.7.0",
-        "jquery": "^3.6.4",
+        "jquery": "^3.7.0",
```
petschki, line 37: I tried that out a while ago, but this breaks the tests. Didn't look into details yet.

petschki (changes requested):

package.json
petschki: jQuery is not the problem. See comment below.

package.json
```diff
@@ -48,18 +48,18 @@
         "@11ty/eleventy": "^2.0.1",
         "@11ty/eleventy-navigation": "^0.3.5",
         "@11ty/eleventy-plugin-syntaxhighlight": "^4.2.0",
-        "@patternslib/dev": "^3.3.5",
+        "@patternslib/dev": "^3.4.0",
         "@types/sinon": "^10.0.14",
         "npm-run-all": "^4.1.5",
         "rimraf": "^4.1.3",
-        "sinon": "^15.0.4"
+        "sinon": "^15.2.0"
```
petschki, line 55: Turned out, that this sinon version has to be in `transformIgnorePattern` in jest config. Now tests are green again.

petschki (approved):

thet:
:tada: 

super!
I had a ton of breaking tests in mockup, glad that these were false flags...

PR #1336, opened 2023-08-04
maint(Docs): fix table in modal.

mauritsvanrees:
I needed to look up the available options for the `modal` pattern, but found that the table with options at the bottom of the `modal` readme is unreadable.  This is because it has invalid MarkDown. See http://plone.github.io/mockup/pat/modal/ or [here on GitHub](https://github.com/plone/mockup/tree/5.1.3/src/pat/modal).

Problem is that the line with underlines for the header contains one column too many: five instead of four.  I have fixed this, and also made the underlines less long.  According to the [MarkDown docs](https://www.markdownguide.org/extended-syntax/#tables) you only need three dashes.

Also, I have removed lots of extraneous white space: the columns do not all have to be the same width in characters.

Tip: when viewing the changes of this PR, add `?w=1` at the end of the url to ignore lines with only white space changes.

Result:

Screenshot 2023-08-04 at 10 49 09


mauritsvanrees:
For easy reviewing, this is the link to the rendered modal readme from this PR:
https://github.com/plone/mockup/blob/f2eb54a09d7743d36622c63024244376a789e07e/src/pat/modal/README.md#configuration

petschki (approved):

petschki:
I have whitespaces permanently hidden here:

Bildschirmfoto 2023-08-04 um 11 51 47

PR #1337, opened 2023-08-05
Fix tinymce link/image modal with current selection

### LINKED ISSUE #1319: Changing internal links/images in TinyMCE throws error
Reproduced on classic.demo.plone.org :

- Edit a Page and insert an internal image to the TinyMCE Editor
- Save the modal
- click the image and open image modal again
- modal is shown but there is a JS error on the console and the relatedItems toolbar is missing:

Bildschirmfoto 2023-05-15 um 12 53 27

Same behavior for internal links.

petschki:
fixes #1319 

NOTE: absolutely no clue why this makes a difference `this.$el` and `$(this.el)` but I changed it back like it was before and it works again. See commit https://github.com/plone/mockup/commit/d8f2daa9f45b6805dd9f35e791437e04bce47267#diff-0372046d8d03fc963bb3d1bedcb015a6eebdcd9cd11cfaf9d604690ba2319434L461-R458

### DIFF
--- a/src/pat/relateditems/relateditems.js
+++ b/src/pat/relateditems/relateditems.js
@@ -458,9 +458,9 @@ export default Base.extend({
 
     selectItem(item) {
         this.emit("selecting");
-        const data = this.$el.select2("data");
+        const data = $(this.el).select2("data");
         data.push(item);
-        this.$el.select2("data", data, true);
+        $(this.el).select2("data", data, true);
 
         if (this.options.recentlyUsed) {
             // add to recently added items


thet (approved):
LGTM, although it's suspicious that jQuery needs to instantiated freshly like this.
But Select2 works that way.
Probably the problem is because when `this.$el` is instantiated Select2 wasn't loaded and the already instantiated jQuery instance doesn't get notice when Select2 is finally loaded.

It's still strange because Select2 is imported on the top of the module and it should actually be already available when pat-relateditems is initialized.

If your fix works fine, let's merge it!

PR #1338, opened 2023-08-08
Fix sticky position of `pat-structure` toolbar when edit toolbar is on top

### LINKED ISSUE #1335: pat-structure and horizontal toolbar
Case:

1. horizontal toolbar
2. a lot of catalog metadatas and columns in the structure table

Problem One:
Toolbar of the Structure Table is Sticky but hidden if i scroll down

Problem Two:
Only with a fixed Taskbar on Windows, the Save Button of the Column Selection UI Panel is not accessible

Screen for better understanding
Bildschirmfoto vom 2023-08-01 09-33-53

petschki:
fixes #1335

### DIFF
--- a/src/pat/structure/structure.scss
+++ b/src/pat/structure/structure.scss
@@ -390,3 +390,7 @@
         }
     }
 }
+
+body.plone-toolbar-top .pat-structure .navbar {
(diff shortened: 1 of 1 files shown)


1letter (approved):
Thanks! LGTM

Issue #1341, opened 2023-09-13
pat-relateditems: Issue with existing value when `Select2` is not imported


frapell:
I found an edge case with the related items widget used in the TinyMCE "Insert Image" or "Insert Link" dialogs. Here are the steps to reproduce:

1. Go to /@@manage-portlets and add a "Static text portlet" (It has to be a portlet)
2. Add some text, highlight it and then click "Insert/edit Link" to choose some content from the site
3. Save the portlet
4. Edit the saved portlet
5. Click the Link
6. Click "Insert/edit Link"

The following error is shown in the dev tools
```
relateditems.js:461 Uncaught TypeError: i(...)(...).select2 is not a function
    at n.selectItem (relateditems.js:461:33)
    at n.updateRelatedItems (links.js:109:27)
    at n.set (links.js:137:14)
    at n.load (links.js:130:18)
    at n.initData (links.js:894:47)
    at n.modalShown (links.js:735:14)
    at HTMLDivElement.<anonymous> (links.js:461:29)
    at HTMLDivElement.dispatch (jquery.js:5135:27)
    at g.handle (jquery.js:4939:28)
    at Object.trigger (jquery.js:8619:12)
```

At this point, clicking "Cancel" or "Insert" keeps the spinning "Loading" and the modal is not closed.
The same behavior is observed with the "Insert Image" dialog.

This doesn't happen if you try this in a regular page, instead of a static text portlet. The reason is because the `Select2` module is imported when the `pat-select2` pattern is initialized for the `Related Items` field. Since the static text portlet doesn't have a `Related Items` field, the init of the pattern is done by the tinymce plugin at
https://github.com/plone/mockup/blob/c9d330a8e06aa0b4e193a9497f3eb5558efc2811/src/pat/tinymce/js/links.js#L95-L110
and the `selectItem` is called before the pattern has been fully initialized

PR #1342, opened 2023-09-13
fix(pat relateditems): Import Select2 module before calling it when not yet initialized

frapell:
This PR fixes gh-1341

### DIFF
--- a/src/pat/tinymce/js/links.js
+++ b/src/pat/tinymce/js/links.js
@@ -352,6 +352,12 @@ tinymce.PluginManager.add("ploneimage", function (editor) {
         onAction: editor.settings.addImageClicked,
         // stateSelector: "img:not([data-mce-object])",
     });
+    editor.ui.registry.addMenuItem("ploneimage", {
+        icon: "image",
+        text: "Insert image",
+        onAction: editor.settings.addImageClicked,
+        // stateSelector: "img:not([data-mce-object])",
+    });
 });
 
 /* register the tinymce plugin */
@@ -359,6 +365,13 @@ tinymce.PluginManager.add("plonelink", function (editor) {
     editor.ui.registry.addButton("plonelink", {
         icon: "link",
         tooltip: "Insert/edit link",
+        shortcut: "Ctrl+K",
+        onAction: editor.settings.addLinkClicked,
+        stateSelector: "a[href]",
+    });
+    editor.ui.registry.addMenuItem("plonelink", {
+        icon: "link",
+        text: "Insert link",
         shortcut: "Ctrl+K",
         onAction: editor.settings.addLinkClicked,
         stateSelector: "a[href]",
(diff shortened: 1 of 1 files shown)


frapell:
@thet Mind taking a look at this if you have some time? My understanding is that the `new` operator doesn't return a promise, so it cannot be awaited... My first approach at fixing this issue was to add an `await` in https://github.com/plone/mockup/blob/c9d330a8e06aa0b4e193a9497f3eb5558efc2811/src/pat/tinymce/js/links.js#L101 but that doesn't work and https://github.com/plone/mockup/blob/c9d330a8e06aa0b4e193a9497f3eb5558efc2811/src/pat/tinymce/js/links.js#L109 gets called with the `select2` module not yet available.

I don't like my solution very much, but I believe a proper solution would require extensive refactoring, and could potentially break in unexpected places...

Thoughts?

src/pat/relateditems/relateditems.js
```diff
@@ -456,7 +456,8 @@ export default Base.extend({
         this.emit("after-browse");
     },
 
-    selectItem(item) {
+    async selectItem(item) {
+        await import("select2");
```
thet, line 460: LGTM

thet:
Hi, I think using the `await_pattern_init` from Patternslib should work.
Import:
```
import events from "@patternslib/patternslib/src/core/events";
```

then:
```
createRelatedItems: async function () {
    [...]
    this.relatedItems = new RelatedItems(this.getEl(), options);
    await events.await_pattern_init(this.relatedItems)
}
```

There should be many examples in Patternslib.

Hope that helps!

I'm happy to review this later - today it's bedtime.

frapell:
@thet Thank you for that! had no idea about the `await_pattern_init` and I think it is a better solution. I force pushed the commit replacing the previous one to avoid excessive commits

frapell:
@thet I included my latest commit into this PR, which is totally unrelated to this fix, but I think it is a nice to have. Right now, it is not possible to add the "Insert Link" or "Insert Image" to the menu, they can only be added as buttons to the toolbar.
I added calls to `addMenuItem` for them (https://www.tiny.cloud/docs/tinymce/6/apis/tinymce.editor.ui.registry/#addMenuItem) and now you can add those actions to the menu:
image

Let me know if you want me to revert it and submit a separate PR for that

thet (approved):
LGTM.

regarding the add image/add link in the insert menu - totally fine for me.
But you have this commit message: "Include ability to add the insert image and insert link in the insert menu". Shouldn't it be: "Include the insert image and insert link in the insert menu"? Because you aren't adding a ability but the menu items itself, right?

frapell:
@thet No, you still need to add `ploneimage` and `plonelink` to the JSON config for the menu in `/@@tinymce-controlpanel`. What happens today without this code change is that if you cannot have the buttons showing up in the menu, even adding those entries to the config.

PR #1346, opened 2023-10-19
TinyMCE 6

petschki:
~~Note: There are renamed menu and toolbar items for the format menu. You have to rename them manually right now.~~

We map the old names now here until the configuration values changed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,8 +40,8 @@
         "js-cookie": "^3.0.5",
         "select2": "git+https://github.com/ivaynberg/select2.git#3.5.4",
         "sortablejs": "^1.15.0",
-        "tinymce": "^5.10.7",
-        "tinymce-i18n": "^22.12.27",
+        "tinymce": "^6.7.0",
+        "tinymce-i18n": "^23.10.9",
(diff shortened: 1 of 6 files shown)


thet:
@petschki can we add aliases for these old values (`formats` and `styleselect`) in Mockup itself? that might be easier than adding an upgrade step and it would be better backwards compatible - you could use Mockup with older versions of Plone too, maybe.

thet (changes requested):
I think we need aliases for the old tinymce toolbar/menu values. Plus, after merging we also need to update the default options for the tinymce control panel to reflect those new values.

Other than that: :tada: :100:

thet (approved):

petschki:
just fixed the tests.

I think semantic versioning would require a `5.2.0` release of this ... should we got with an alpha first? and `plone.staticresources = 2.2.0` too ...

thet (approved):

thet:
Yes for 5.2.0.
An alpha version might be good to check if the integration works properly with Plone.

MrTango:
great job! thx

PR #1357, opened 2023-12-12
Update dependencies [5.1.x]

petschki:


thet (changes requested):
Please also include Patternslib 9.10.0. It has plenty of fixes for an otherwise partly broken pat-inject/autosubmit which broke after 9.5.0.

thet (approved):

PR #1370, opened 2024-02-26
Fix duplicated image node when updateing already inserted image. [5.1.x]

petschki:
fixes https://github.com/plone/Products.CMFPlone/issues/3910

### DIFF
--- a/src/pat/tinymce/js/links.js
+++ b/src/pat/tinymce/js/links.js
@@ -681,8 +681,10 @@ export default Base.extend({
         self.tiny.selection.setRng(self.rng);
         var cssclasses = [
             "image-richtext",
-            self.$align.val(),
         ];
+        if(self.$align.val()) {
+            cssclasses.push(self.$align.val());
+        }
         if(self.linkType !== "externalImage"){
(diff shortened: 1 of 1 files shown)


pbauer:
How do I test that? I forgot how to get Plone to use the raw js instead of the minified bundle.

petschki:
1. checkout `mockup`
2. run `yarn install && yarn start:webpack` inside the package
3. in the resource controlpanel go to the plone bundle and use `http://localhost:8000/bundle.min.js` as javascript resource

pbauer:
I tested https://github.com/plone/plone.staticresources/pull/323 with 6.1 and it worked. 
`yarn start:webpack` didi not work:

```
mockup git:(5.1.x) $ yarn start:webpack
command not found: webpack`
```

pbauer (approved):

petschki:
> 
> `yarn start:webpack` did not work:
> 

hm ... strange. maybe this is because of yarn version ... mockup only runs in `yarn < 2` 

I'll prepare releases now for the fix.

PR #1376, opened 2024-03-05
Feature/help plugin

rber474:
After issue https://github.com/plone/plone.base/issues/41, a [PR](https://github.com/plone/plone.base/pull/59) was requested in order to make `help` plugin available throught TinyMCE controlpanel.

Checking this plugin option in would display a toolbar or menu options to show the help view in the editor but an exception was rised because of it could find the resources for navkey translation pages.

This aproach fixes that issue and refactors the code. Now help is displayed and translations are properly shown.

1. Uses original plugin code but refactoring it to improve readability.
2. Includes all navkey translation resources inside the mockup tinymce folder to solve the load error.
3. Loads `help.js` in `tinymce-implementation.js`

I am not sure if this is the best way to solve the behaviour but I've implemented it in a production site and fits our needs.

Please, review it. Any comments are welcome.

@rber474


.husky/commit-msg
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-yarn commitlint --edit 
+# yarn commitlint --edit
```
gforcada, line 4: That's probably a leftover?

rber474: hmmm... nothing I recall to have changed. Will restore and commit original state

rber474 (commented):
Fixed commented line and a couple of errors.

petschki:
Just wondering why this is needed at all, because the help plugin is shipped with the tinymce package ... if the plugin is activated in the controlpanel (which got added in `plone.base` I think?) the pattern tries to import it here: https://github.com/plone/mockup/blob/master/src/pat/tinymce/tinymce--implementation.js#L162 ... to see the `log.debug` messages you need to add `?loglevel=DEBUG` to the browser address then the messages are displayed in the console.

petschki:
@rber474 I'll close this one. If you have any further problems with the help plugin please open a new PR.

PR #1377, opened 2024-03-14
`pat-contentbrowser` implementation

petschki:
This is the `pat-relateditems` replacement using a miller column content browser.

### Feature completeness:

- [x] Multiple `pat-contenbrowser` occurrences on a single page
- [x] TinyMCE `LinkModal` implementation
- [x] filtering current level list
- [x] `maximumSelectionSize`
- [x] `pat-upload` to current level
- [x] selectable levels
- [x] batchable levels
- [x] drag&drop sortable items
- [x] `selectableTypes`
- [x] translations
- [x] multiple items select in column browser
- [x] keyboard navigation
- [x] TESTS
- [ ] favorites menu
- [ ] recently used menu
- [ ] use @plone/registry to allow overriding components

### Test dependencies:

- https://github.com/plone/plone.app.z3cform/pull/197
- https://github.com/plone/plone.app.relationfield/pull/51
- then run `yarn start:webpack` 
- then use `http://localhost:8000/bundle.min.js` as `plone` bundle in registry controlpanel

Regarding `search mode`: 
we decided to not implement this here. The `pat-select2` pattern is better for this usecase.
UPDATE: Since this should be a feature complete replacement for `pat-relateditems`, the `search` mode is available too.

#### TinyMCE Link/Image

https://github.com/plone/mockup/assets/511761/abb3ed2c-d0b6-48c6-8945-8429dca775aa

#### Related items selection and resorting

https://github.com/plone/mockup/assets/511761/71c592e1-5f89-451c-bc9e-725a61185077

#### Multiselection with shift or ctrl/cmd key

https://github.com/plone/mockup/assets/511761/af845c2e-6ab6-4934-94d8-79e17390ffb1

#### Keyword filtering

https://github.com/plone/mockup/assets/511761/c7c387bd-22d7-42a5-8a52-b5b26c4c1b47

#### Uploading

https://github.com/plone/mockup/assets/511761/5bc1e550-7364-4728-9363-88a11014496f


thet:
looking at the video, great!
how can this be tested?
are there branches in plone repos using contentbrowser instead of relateditems?
or should we maybe provide an alias "relateditems" for "contentbrowser"?
i think this also needs some integration in docs.

petschki:
This can be tested by using the mentioned branches of `p.a.z3cform` and `p.a.relationfield` and the webpack devel resources in the resource registry. The relationfield uses the contentbrowser widget already. 

Updateing all the core packages which use `RelatedItemsFieldWidet` isn't that hard but there are several addons which have to be updated too ... I'm not sure if its a problem when we "alias" `RelatedItemsFieldWidget` to `ContentBrowserFieldWidget` under the hood. Maybe that does work for the addons. I can give this a test ...

All in all the main benefit from removing `pat-relateditems` is that we can get rid of the patched select2 `3.5.x` version.

thet:
Cool! Sorry I didn't read the ticket description carefully enough.

petschki:
@MrTango @thet I tried to figure out the `jest` setup for `svelte` apps using this link https://www.roboleary.net/2021/11/18/svelte-app-testing-jest.html ... now I get the following message:

```
Jest is being called in CJS mode. You must use ESM mode in Svelte 4+
```

Maybe someone of you can give me a hint how to setup the testing environment?

thet (approved):
@MrTango @petschki great work so far!

All in all this LGTM so far.

Some remarks though:

- Sorting doesn't seem to work. I see there is some sorting code in `SelectedItems.svelte` but for me adding multiple items and trying to sort just doesn't work.

Some feature requests, which are irrelevant to this PR but anyhow:

- Multiple selection would be nice. Currently you can only select one by one.

- I'd love some kind of keyboard navigation would be great. Like Arrow navigation, enter to move in a lower directory and setting the focus in that directory, enter to select a node.

- Using plone.restapi instead of custom endpoints. IMO this is something we really should go for. Why would we have custom JSON endpoints implemented when we can have everything and more through plone.restapi? But for the first iteration depending on the custom JSON endpoints like pat-relateditems is perfectly OK.

petschki:
@thet Thanks for your review!

About sorting: this only gets initialized on loading the first time and doesn't work during "first time" item selection indeed. (https://github.com/plone/mockup/pull/1377/files#diff-4b241e899bfbe65a8c35ffadf58f567fea3d4df013c56306db87d1517e10798dR83) this is a small change and I will  fix it asap. I'm not sure if we should use `pat-sortable` for this usecase.

Features:

- Multiselection is definitely on my list to implement.
- keyboard navigation too
- and restapi: yes of course (!) but this needs some more thoughts about "inplace" replacement of `pat-relateditems` and its options. We often use relateditems pattern as `mode="search"` with some additional index like `portal_type=["Image", "File"]` ...  IMHO this should work here too ...

MrTango:
regarding keyboard navigation, yes definitely, i think i had some key's already implemented like ESC to close the browser, but i had to disabled some stuff, when i did update to Svelte 4. Not sure if is enabled already. That was a plugin.
But yeah, i want all the keyboard navigation we have in relateditems widget now and more.

regarding RestApi, here from today's classic UI meeting:

- should we use `plone.restapi` for the JSON endpoints?
    - talking with @mrTango, @jensens, @petschki we should do not because it should be possible to use Plone classic UI without restapi (jens mentioned circular dependency problems.)
    - the vocabulary end point in Plone core has all features we need.

MrTango:
and also multiselection is already on the list, even if we didn't had it before in relateditems.

MrTango:
cool, great work.
Have a look at https://github.com/metonym/svelte-keydown.
It's currently disabled i think, because i had problems when updating to Svelte 4.
But the repository was updated 2 months ago, so just try to enable it again.
There should be some comment out examples for closing the content browser with ESC and maybe others.
But navigating was not fished yet, i think.

MrTango:
here an example how the closing of the browser done
```
<!-- <Keydown paused={!$showContentBrowser} on:Escape={cancelSelection} /> -->
```

thet:
I fixed the tests in https://github.com/plone/mockup/pull/1383

thet (approved):

src/pat/tinymce/templates/upload.xml
```diff
@@ -1,7 +1,7 @@
 <div>
-  <div class="linkModal">
+  <div class="linkModal22">
     <h1><%- uploadHeading %></h1>
-    <input type="text" name="upload" style="display:none" />
-    <div class="uploadify-me"></div>
+    <input type="text" name="upload22" style="display:none" />
+    <div class="uploadify-me22"></div>
```
thet, line 5: Why do the classes in here have the suffix `22`?

petschki: @MrTango was this intentionally or a mistake? see https://github.com/plone/mockup/commit/93d2876114f4feb2f6a340b8003f3fbb7c5716f3#diff-34976cb7c710995cf0a4bf63c313426e4921ec7386d82e9c42f85fd489a65a66R2

yurj:
Hi! can I ask if this browser can select the Plone Site object? The `Plone Site` is a content type and should be selectable by the internal browser.

petschki:
> Hi! can I ask if this browser can select the Plone Site object? The `Plone Site` is a content type and should be selectable by the internal browser.

Thanks for the tip. The Portal Content is now selectable (as long its not disabled via `selectableTypes`). I also polished the path label and removed the physical portal path from it.

Bildschirmfoto 2024-06-18 um 09 32 35

petschki:
We have new robottests and fixed the existing ones for the contentbrowser implementation.
This tests the implementation in tinymce and relationfield for `relatedItems`.
See https://github.com/plone/Products.CMFPlone/pull/4008 ... thanks to @1letter  !

thet:
@petschki If you think this is ready, just go ahead and merge :)

petschki:
OK ... I'll merge the current state and release another alpha of that.

Next week I'll work on :

- favorites menu
- recently used menu

and eventually :

- use @plone/registry to allow overriding components

Issue #1379, opened 2024-05-08
Bugs when editing "Selected dates" in pat-recurrence


laulaz:
This is a grouped issue because 2 bugs come from the exact same date formatting problem.

There are (at least) 3 dates format used in recurrence edition : the first one coming from `plone.formwidget.recurrence` through `@@json_recurrence`(`20240508T000000`), the one used in the pattern (`2024-05-08T000000`) and the one used when adding date through the date field (`2024-05-08`).

This is causing issues in dates edition of recurrences.

### Bug 1 : you can add several times the same date

https://github.com/plone/mockup/assets/1101273/5fea60e7-a613-4487-b78a-1eaf147baeb1

**Reason :** the date added is referenced by the input field in format `2024-05-08` and the dates stored in the field are in format `2024-05-08T000000`.
Thus, this line never gets the comparison right : https://github.com/plone/mockup/blob/2058cfb695ba6311d680b27a6d6f650dbf87ba62/src/pat/recurrence/recurrence.js#L698
 <br/>

### Bug 2 : the wrong date is removed when you remove a date

https://github.com/plone/mockup/assets/1101273/22018353-3660-4c1f-960c-b701743238d6

**Reason :** the date you remove is referenced (by `<a date="...">`) in format `20240508T000000` and the dates stored in the field are in format `2024-05-08T000000`.
Thus, these lines don't do the slicing correctly and remove the last date as `
$.inArray(this.attributes.date.value, textarea["ical"].RDATE)` will always return `-1` :
https://github.com/plone/mockup/blob/2058cfb695ba6311d680b27a6d6f650dbf87ba62/src/pat/recurrence/recurrence.js#L678-L681

petschki:
I think this can be closed, right?

PR #1380, opened 2024-05-08
Fix comparisons between different dates formats

laulaz:
This refs #1379


petschki (changes requested):
Hi, thanks for this! I could test your branch and can confirm, that removing a date works now correctly.

But adding a date which already exists is still allowed whether its in the RRULE or its added manually. I just quickly tried your screencasts but could not look into the code.

petschki:
Took a closer look and now I understand: The added dates aren't checked against an RRULE but only for already added custom RDATE entries. This is ok for me because its too much overhead to check the custom date in the backend if it already exists in the RRULE. This would be much easier if we switch to a JS rrule library to check the custom dates there ... see #1210 ... but for now I'll merge. @laulaz could you backport this also to branch `5.1.x` please?

petschki (approved):

PR #1382, opened 2024-05-16
Translate Close text as aria-label

### LINKED ISSUE #1381: Modal "close" button's aria-label is not translated
https://github.com/plone/mockup/blob/master/src/pat/modal/modal.js#L61

erral:
Fixes #1381 

@petschki can you check if this syntax is correct?

### DIFF
--- a/src/pat/modal/modal.js
+++ b/src/pat/modal/modal.js
@@ -4,7 +4,7 @@ import _ from "underscore";
 import Backdrop from "../backdrop/backdrop";
 import registry from "@patternslib/patternslib/src/core/registry";
 import utils from "../../core/utils";
-import _t from "../../core/i18n";
+import _t from "../../core/i18n-wrapper";
 
 import "jquery-form";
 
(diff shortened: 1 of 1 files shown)


src/pat/modal/modal.js
```diff
@@ -58,7 +58,7 @@ export default Base.extend({
                 '    <div class="<%= options.classModal %>" role="document">' +
                 '      <div class="<%= options.classHeaderName %>">' +
                 '        <% if (title) { %><h5 class="modal-title" id="modal-title" tabindex="0"><%= title %></h5><% } %>' +
-                '        <button type="button" class="btn-close modal-close" aria-label="Close"></button>' +
+                '        <button type="button" class="btn-close modal-close" aria-label="<%= _t("Close") %>"></button>' +
```
petschki, line 61: Unfortunately this doesn't work. You need to define a property, for example `tplObject.closeLabel` with the translated label and use it in the template like `<%= closeLabel %>`. See https://github.com/plone/mockup/blob/23f4d61cff105e946fea7feb5202da12bdce6ffe/src/pat/modal/modal.js#L392-L399 how its done with the other vars.

petschki (changes requested):

petschki (approved):

erral:
Sorry for leaving this unattended :(

PR #1384, opened 2024-06-12
Backport `template` plugin from TinyMCE 6

petschki:


1letter (approved):
i tested this here in my local sites and i see no errors. i can insert template like in version tinymce < 7.

thet (commented):
Does the TinyMCE license allow us to use the TinyMCE 6 plugin in TinyMCE 7?
If yes, I'm fine with merging this.

petschki:
TinyMCE lincense is GPLv2+ ... since the "old" `template` plugin is distributed as "open source template" I assume it's ok to use the code ... we have to maintain it ourselfes anyway if it doesn't work anymore in a future tiny release. https://www.tiny.cloud/docs/tinymce/6/template/

thet (approved):

MrTango (approved):

PR #1391, opened 2024-08-21
`pat-tinymce` save value on 'change' event to fix validation

### LINKED ISSUE #1390: required `pat-tinymce` leads to unsaveable forms.
### Steps to reproduce:

1. create a `required` `schema.Text` field with `pat-tinymce` enabled (eg. `plone.app.z3cform.widgets.richtext.RichTextWidget`)
2. go to the editform and save it empty
3. the backend validation comes back with error
4. fill out value in tiny
5. save button is still deactivated, because changeing the tiny value doesn't trigger `pat-validation` to re-validate -> you cannot save form.

This is due to the new z3cform textarea implementation which introduces the required parameter for the richtext field textarea now. In Plone 6.0 the plone.app.textfield markup ignored this parameter and only validated required in the backend.

So thats further why pat-validation deactivates the save button now if a required richtextfield has no input.

### Possible solution

try to refactor pat-tinymce to save the value on blur event which would trigger pat-validation again and should fix this problem.

petschki:
fixes #1390

### DIFF
--- a/src/pat/tinymce/tinymce--implementation.js
+++ b/src/pat/tinymce/tinymce--implementation.js
@@ -1,5 +1,6 @@
 import $ from "jquery";
 import I18n from "../../core/i18n";
+import events from "@patternslib/patternslib/src/core/events";
 import logger from "@patternslib/patternslib/src/core/logging";
 import _t from "../../core/i18n-wrapper";
 import utils from "../../core/utils";
@@ -10,6 +11,7 @@ let LinkModal = null;
 
 export default class TinyMCE {
     constructor(el, options) {
+        this.el = el;
         this.$el = $(el);
         this.options = options;
     }
@@ -145,8 +147,8 @@ export default class TinyMCE {
         if (theme && theme == 'dark') {
             css = 'oxide-dark';
         }
-        import (`tinymce/skins/ui/${css}/content.css`);
-        import (`tinymce/skins/ui/${css}/skin.css`);
+        import(`tinymce/skins/ui/${css}/content.css`);
+        import(`tinymce/skins/ui/${css}/skin.css`);
 
         const tinymce = (await import("tinymce/tinymce")).default;
         await import("tinymce/models/dom");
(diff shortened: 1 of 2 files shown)


1letter:
i can test it tomorrow, sorry forthe delay

1letter (approved):
I can not say anything about the JS code, but in my 3 Testsystems with Plone 6.1a4 it'works like expected. Thanks!

src/pat/tinymce/tinymce--implementation.js
```diff
@@ -232,6 +234,14 @@ export default class TinyMCE {
                     ]);
                 },
             });
+            // handle 'change' event to ensure correct validation (eg. required textfield)
+            // eslint-disable-next-line no-unused-vars
+            editor.on("change", (e) => {
+                // setting tiny content manually
+                this.el.value = editor.getContent();
+                // dispatch "change" event for pat-validation
+                this.el.dispatchEvent(new Event("change"));
```
thet, line 243: Please use the `change_event` handler from Patternslib because that one bubbles like the original `change` event. Something one often forgets (unless you explicitly don't want to let it bubble).

Do something like:
```
import events from "@patternslib/patternslib/src/core/events";

// ...


                this.el.dispatchEvent(events.change_event());
```

thet (changes requested):
LGTM! I also like the simplification, not doing different things for inline or iframe based TinyMCEs. That should just work for overlays too.

We have to keep in mind, that this saved back at every keystroke, right?
That's done similarly with other editors too, e.g. the pat-tiptap IIRC, which I maintain. However, this could be a potential performance problem.

One change requested, though.

petschki:
TinyMCE fires its change event when you leave the editor (blur) see https://www.tiny.cloud/docs/tinymce/latest/events/ 

I'll change the eventhandler call next week as suggested.

petschki:
I'm using the patternslib event utils now. I also tested this in overlays. Mosaic uses Tiny in the options overlay ... need this regularly 😉

PR #1393, opened 2024-09-11
re-enable `pat-structureupdater`

petschki:
see https://github.com/plone/Products.CMFPlone/issues/4011

The configured selectors needs to be fixed here https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/patterns/settings.py#L41-L56


thet (approved):
I was chewing a bit if we should instead rather dispatch a `pat-update` event from the structure pattern and use `pat-inject` to reload the title and description.
but the title/description info is already available via the contentbrowser JSON api endpoints and passed in the event - so this is more efficient.
Also, it's a fix and not a breaking change :)

So LGTm!

PR #1394, opened 2024-09-23
SASS 1.79 compatibility

thet:
Begin work on SASS 1.79 compatibility.

Ref:
https://github.com/twbs/bootstrap/issues/40849
https://github.com/twbs/bootstrap/pull/40865
https://github.com/twbs/bootstrap/pull/40864

IIRC (worked on that some days ago) this is not really silencing the deprecation warnings from SASS.

Probably pinning SASS to something before 1.79 is better (`resolveDependencies`).

### DIFF
--- a/src/pat/relateditems/relateditems.scss
+++ b/src/pat/relateditems/relateditems.scss
@@ -1,3 +1,5 @@
+@use 'sass:color';
+
 .pat-relateditems-container {
     position: relative;
     display: inline-block;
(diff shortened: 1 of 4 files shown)


petschki (approved):

thet (commented):
The important change here is from @petschki which removes the dependency on Bootstrap in structure.scss.
LGTM, will merge.

PR #1396, opened 2024-09-27
`pat-contentbrowser` component registry

petschki:
This registers `SelectedItem` component with `@plone/registry` to make it overridable.

### DIFF
--- a/src/pat/contentbrowser/README.md
+++ b/src/pat/contentbrowser/README.md
@@ -29,6 +29,7 @@ Show a widget to select items in an offcanvas miller-column browser.
 |       recentlyUsed         | boolean |                  false                  |                                                           Show the recently used items dropdown.                                                          |
 |      recentlyUsedKey       | integer |                                         |                     Storage key for saving the recently used items. This is generated with fieldname and username in the patternoptions.                  |
 |    recentlyUsedMaxItems    | integer |                    20                   |                                              Maximum items to keep in recently used list. 0: no restriction.                                              |
+|     customComponentKeys    |  dict   |                    {}                   |                                           Register custom components. Currently only "SelectedItem" implemented                                           |
 
 
 ## Default
@@ -81,3 +82,44 @@ Show a widget to select items in an offcanvas miller-column browser.
     data-pat-contentbrowser='{"selectableTypes": ["Image", "File"], "vocabularyUrl": "contentbrowser-test.json", "upload": true}'
 />
 ```
+
+## Register custom component
+
+Currently only for `SelectedItem` component available.
+
+```html
+<input
+    type="text"
+    class="pat-contentbrowser"
+    data-pat-contentbrowser='{
+        "customComponentKeys": {
+            "SelectedItem": "pat-contentbrowser.myfield.MySelectedItemComponent",
+        },
+    }'
+/>
+```
+
+Copy the existing component `src/SelectedItem.svelte` to your addon, customize it and register it in your JS bundle as follows:
+
+```javascript
+...
+import plone_registry from "@plone/registry";
+...
+
+async function register_selecteditem_component() {
+    // we register our component to a custom keyname,
+    // which later can be used for pattern_options
+    const SelectedImages = (await import("./MySelectedItemComponent.svelte")).default;
+    plone_registry.registerComponent({
+        name: "pat-contentbrowser.myfield.MySelectedItemComponent",
+        component: SelectedImages,
+    });
+}
+register_selecteditem_component();
+
+...
+
+```
+
+Note: this needs the `svelte-loader` plugin in your webpack.config.js ... see mockups webpack config for info.
+
--- a/src/pat/contentbrowser/src/RecentlyUsed.svelte
+++ b/src/pat/contentbrowser/src/RecentlyUsed.svelte
@@ -30,7 +30,7 @@
             {#each items.reverse() as recentlyUsed}
                 <li>
                     <a
-                        href={recentlyUsed.getURL}
+                        href={recentlyUsed.getURL || "#"}
                         on:click|preventDefault={() => select(recentlyUsed)}
                         class="dropdown-item"
                     >
--- a/src/pat/contentbrowser/src/SelectedItems.svelte
+++ b/src/pat/contentbrowser/src/SelectedItems.svelte
@@ -1,13 +1,17 @@
 <script>
-    import { getContext, onMount } from "svelte";
-    import { flip } from "svelte/animate";
+    import { getContext, onMount, setContext } from "svelte";
     import { get_items_from_uids, resolveIcon } from "./utils.js";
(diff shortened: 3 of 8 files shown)


petschki:
This solution provides a general approach to override the component in general with `@plone/registry` but we also should provide a way, so that an integrator is able to do this only for the widges he wants. In `relateditems` it was possible with `pattern_options` which could be overriden with autoform schemahints, but here I do not (yet) have an idea how to bring this together with `@plone/registry` ...

@MrTango @thet you might have an idea on this?

MrTango:
So if we combine the lookup from the registry, with pattern_options for each component we want to make customizable, we can change anything, even on per field level.

For example, if we want this for SelectedItem, the default name for the component is `pat-contentbrowser.SelectedItem`, but via a pattern_option `selected_item_component` we could say that we want `pat-contentbrowser.SelectedItemAdvanced` to be used and register it also in the registry. And even that, one could override later in a different package via registry.

petschki:
Historical: you could provide either a `selectionTemplateSelector` option, which had a CSS selector to a html snippet or you put the full template code (underscore XML template) inline with `selectionTempate` option. This is then applied with the `formatSelection` method from `select2` https://github.com/plone/mockup/blob/master/src/pat/relateditems/relateditems.js#L531 ... and there were more templates to override: https://github.com/plone/mockup/blob/master/src/pat/relateditems/relateditems.js#L507-L513 ...

I like the idea of providing custom registry keys to the pattern options. I could think of a dictionary, since there could be more overridable components in the future:

```
data-pat-contentbrowser=`{
    "components": {
        "selectedItem": "pat-contentbrowser.MyAdvancedSelectedItem",
        "browserItemPreview": "xyz",
        ...
}'
```

thet:
@petschki this is an interesting approach. But I think this misses the use case a bit.

With the previous pat-relateditems approach you could change the selected item template per related items widget instance. Now you can only change the component for all reference browser instances.

What I think would be really useful is to use a different selected item component based on the context. As equivalent in Plone: a different browser view based on the implemented interface of the context.

It looks like that the @plone/registry does support this via the `dependencies` option (they should have called it `context` or `contexts`):

https://plone-registry.readthedocs.io/how-to-guides/register-and-retrieve-components.html

src/pat/contentbrowser/src/RecentlyUsed.svelte
```diff
@@ -30,7 +30,7 @@
             {#each items.reverse() as recentlyUsed}
                 <li>
                     <a
-                        href={recentlyUsed.getURL}
+                        href={recentlyUsed.getURL || "#"}
```
thet, line 33: different scope than the component registry work here, right?

src/pat/contentbrowser/src/SelectedItems.svelte
```diff
@@ -85,6 +92,10 @@
         selectedItemsNode.dispatchEvent(events.change_event());
     }
 
+    function LoadSelectedItemComponent(node, props) {
+        const component = new RegisteredSelectedItem.component({target: node, props: props});
```
thet, line 96:
```suggestion
        const component = plone_registry.getComponent({
            name: "pat-contentbrowser.SelectedItem",
            dependencies: [props.item["@type"]]
        }).component;
        return component({target: node, props: props});
```
NOTE: untested pseudo code

src/pat/contentbrowser/src/SelectedItems.svelte
```diff
@@ -21,6 +21,9 @@
     // showContentBrowser reactive state
     const showContentBrowser = getContext("showContentBrowser");
 
+    // get selectedItem component from registry
+    const RegisteredSelectedItem = plone_registry.getComponent("pat-contentbrowser.SelectedItem");
+
```
thet, line 26:
```suggestion
```
Replaced by below...

thet (dismissed):
See my comment here: https://github.com/plone/mockup/pull/1396#issuecomment-2514527410

src/pat/contentbrowser/src/RecentlyUsed.svelte
petschki: `RecentlyUsed` is currently not registered ... this is only a fix unrelated to this PR.

thet:
According to the @plone/registry docs the `dependencies` list allows to define one component for multiple dependencies.
I don't more about the details, but the example shows that `@type` is used.
I'd be OK having a per-type lookup - as long as it doesn't add a significant performance impact. It would be another improvement over the previous situation. That way you could for example render the selected items for different content types differently - an image has different layout requirements than a news item or a location venue or a event.

MrTango:
I prefer the version where we control it in the widget, then we can have multiple veriantion in the same content type for different fields. @thet the registry like we use it now is globally. But if we allow chnaging the name of the component to use via patterns_options and push this into the svelte app via props, we are very flexible on per widget level.
I don't know how the dependency parametr in plone/registry works, but it sounds like something different than what we want, if it's not on field/widget level.

petschki:
I followed the idea of @MrTango and added a new parameter `componentRegistryKeys` https://github.com/plone/mockup/pull/1396/commits/bbbf59d42a50464a67102a6acf22f8989c85349d

This parameter (a dictionary to make other components also customizable in the future) can be customized with `pattern_options` to provide a custom key with a customized component. I've implemented this in an addon from us:

Custom Component: https://github.com/collective/collective.behavior.relatedmedia/pull/18/commits/8ddfaae6d021a69ca9767ff722a58d2d0a9017f1
Pattern options: https://github.com/collective/collective.behavior.relatedmedia/pull/18/commits/8e168aba41903b2e07810280765d8cde37dc820d

And it works 🎉 

@thet the `dependencies` approach could be additinally implemented to further customize "per type" item. But that is out of scope of this PR I think.

src/pat/contentbrowser/README.md
```diff
@@ -81,3 +82,44 @@ Show a widget to select items in an offcanvas miller-column browser.
...
+
+Currently only for `SelectedItem` component available.
+
+```html
+<input
+    type="text"
+    class="pat-contentbrowser"
+    data-pat-contentbrowser='{
+        "customComponentKeys": {
+            "SelectedItem": "pat-contentbrowser.myfield.MySelectedItemComponent",
```
thet, line 96: Regarding the custom component keys  - isn't `selectedItem` the only component key which is supported yet?
I'm a bit restrained to use JSON structures as pattern options - especially JSON objects. Until now I had the impression that JSON objects prevent us from using the Patterns options parser and thus prevent us currently from using the BasePattern. But your PR here shows that this is not entirely true - You're using both the Patterns Parser and BasePattern here. However, at least the `customComponentKeys` is currently not formally defined in the Patterns Parser.
Probably all is good, but I'd have maybe made a flat single option out for the selectedItem component. Something like `selectedItemComponent`.

PR #1406, opened 2024-11-13
Fix focus trap when modal form is reloaded with validation errors

petschki:
fixes https://github.com/plone/Products.CMFPlone/issues/3969

### DIFF
--- a/src/pat/modal/modal.js
+++ b/src/pat/modal/modal.js
@@ -213,10 +213,10 @@ export default Base.extend({
                     $form.append(
                         $(
                             '<input type="hidden" name="' +
-                                $action.attr("name") +
-                                '" value="' +
-                                $action.attr("value") +
-                                '" />'
+                            $action.attr("name") +
+                            '" value="' +
+                            $action.attr("value") +
+                            '" />'
                         )
                     );
                 }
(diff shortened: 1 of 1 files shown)


src/pat/modal/modal.js
```diff
@@ -767,68 +761,64 @@ export default Base.extend({
...
-                inputs.push(inputsBody[i]);
-            }
-        }
-        for (var j = 0; j < inputsFooter.length; j++) {
-            if ($(inputsFooter[j]).is(":visible")) {
-                inputs.push(inputsFooter[j]);
+
+        for (const el of [...inputsBody, ...inputsFooter]) {
+            const style = window.getComputedStyle(el);
+            if (style.display !== 'none' && style.visibility !== 'hidden' && style.opacity !== '0') {
```
thet, line 775: Instead of this check, consider using `is_visible` from `patternslib/src/core/dom`

thet (changes requested):

src/pat/modal/modal.js
```diff
@@ -767,68 +761,64 @@ export default Base.extend({
...
-                ) {
+            modal_el.addEventListener("click", (e) => {
+                if (!e.target.closest(`.${self.options.templateOptions.classModal}`)) {
                     self.hide();
                 }
             });
         }
 
-        self.$modal.find(".modal-title").focus();
+        if (firstInput && ["INPUT", "SELECT"].includes(firstInput.nodeName)) {
```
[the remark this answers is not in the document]
petschki: ops ... forgot it 😄 

done 👍🏼

src/pat/modal/modal.js
petschki: this might be interresting for checking visibility in patternslib: https://developer.mozilla.org/en-US/docs/Web/API/Element/checkVisibility

thet: Oh nice! I love how the web is evolving.

thet (approved):

PR #1411, opened 2024-11-18
pat contentbrowser: ui improvements for select button in toolbar

1letter:
- use the title of content object as label of the select button
- enable a bs tooltip with the path info of selected contentobject

the old behavior was "display the path", now "display the title" and show the path in a tooltip

### DIFF
--- a/src/pat/contentbrowser/src/ContentBrowser.svelte
+++ b/src/pat/contentbrowser/src/ContentBrowser.svelte
@@ -342,13 +342,6 @@
         }, 500);
     }
 
-    function initToolTips(){
-        // init the tooltips for select buttons in contentbrowser
-        const tooltipTriggerList = document.querySelectorAll('[data-bs-toggle="tooltip"]')
-        const tooltipList = [...tooltipTriggerList].map(tooltipTriggerEl => new bootstrap.Tooltip(tooltipTriggerEl))
-        console.log(tooltipTriggerList, tooltipList)
-    }
-
     $: {
         if ($showContentBrowser) {
             contentItems.get({ path: $currentPath });
@@ -416,7 +409,6 @@
                         <div
                             class="levelColumn{i % 2 == 0 ? ' odd' : ' even'}"
                             in:fly|local={{ duration: 300 }}
-                            use:initToolTips
                         >
                             <div class="levelToolbar">
                                 {#if i == 0 && $config.mode == "browse"}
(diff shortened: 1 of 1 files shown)


1letter:
@petschki @MrTango let me know what is your opinion for the requested solution?

1letter:
I have shown this solution to our power users/editors. They want a cleaner solution. ‘Less is more’ - just the “Select” button without text, but that would mean the button for the selection should be in the same column above, not in the next right column up. Perhaps is the solution of this PR overloaded? Are we to nerdy in our Plone world ? I convert this PR to a Draft PR.

src/pat/contentbrowser/src/ContentBrowser.svelte
```diff
@@ -426,11 +434,15 @@
                                 {#if level.selectable}
                                     <button
                                         class="btn btn-primary btn-xs"
+                                        data-bs-toggle="tooltip"
+                                        data-bs-title="{level.displayPath}"
+                                        data-bs-placement="top"
+                                        data-bs-custom-class="contentbrowser-tooltip"
```
thet, line 440: wouldn't a `title` attribute be enough for showing a tooltip by browser mechanisms?
We could also augment that later using patternslib or something, if the browser tooltip isn't enough.

Background: I overheard that Twitter Bootstrap is in declining development state. And I get always a bit nervous when we add another dependency to Bootstrap. Also, the browser tooltip mechanism via the `title` attribute is fine enough in most cases.

thet, line 440:
```suggestion
                                        title="{level.displayPath}"
```

thet (changes requested):
1) I'd not use bootstrap in here but use the browser's `title` mechanism to show tooltips. See me suggestion in code.

2) The commit message could be a bit more verbose about what your changes are about.

src/pat/contentbrowser/src/ContentBrowser.svelte
petschki: I second to not use Bootstrap here.

1letter: I like the idea of simplifying it. I will change it.

MrTango:
I'm wondering if it make sense at all, that we allow selection of parents. From the user point of view, it seems cleaner that you can only select from the current context, not above. The path element would be than only path elements and not buttons. From the UX, i could see, that we move the select button back inside the last column, where it is either an item (Page, Event, Image...) or another Folder. The Folder of course contains the list of items, but it could also have a select button above "select this container".

Issue #1412, opened 2024-11-30
Avoid showing the message `Page: 1 of 0 shown`


devdanzin:
In https://github.com/plone/Products.CMFPlone/issues/3591, it's reported that the message `Page: 1 of 0 shown` is displayed when there are no pages. I think I've found the issue in mockup, here:

https://github.com/plone/mockup/blob/51e23105cb9ac799cf88914e202398eb12c576b0/src/pat/structure/templates/paging.xml#L46-L51

I think the fix should be here, checking that response.total is zero and setting this.state.totalPages to 1:

https://github.com/plone/mockup/blob/51e23105cb9ac799cf88914e202398eb12c576b0/src/pat/structure/js/collections/result.js#L70-L73

Does this make sense? Should I send a PR?

PR #1413, opened 2024-12-01
Show 'Page 1 of 1' instead of 'Page 1 of 0'.

devdanzin:
This trivial PR intends to fix https://github.com/plone/mockup/issues/1412 and https://github.com/plone/Products.CMFPlone/issues/3591 by checking that `totalPages` is zero and setting it to one.

### DIFF
--- a/src/pat/structure/templates/paging.xml
+++ b/src/pat/structure/templates/paging.xml
@@ -44,8 +44,10 @@
   </ul>
 
   <div class="pagination-info">
+    <% if(totalPages){ %>
     <%- _t("Page:") %> <span class="current"><%- currentPage %></span>
     <%- _t("of") %>
     <span class="total"><%- totalPages %></span>
           <%- _t("shown") %>
+    <% } %>
   </div>


src/pat/structure/js/collections/result.js
```diff
@@ -70,6 +70,7 @@ export default Backbone.PageableCollection.extend({
     parseState: function (response, queryParams, state, options) {  // eslint-disable-line
         this.state.totalRecords = response.total;
         this.state.totalPages = Math.ceil(response.total / state.pageSize);
+        this.state.totalPages = this.state.totalPages === 0 ? 1 : this.state.totalPages;
```
thet, line 73:
```suggestion
        // Calculate the number of pages and set it to "1" if there are no results.
        this.state.totalPages = Math.ceil(response.total / state.pageSize) || 1;
```

thet (changes requested):
Hi, thanks for the PR!

Some remarks though:

1) We're using automatic changelog generation based on commit messages using release-it, conventional-changelog and commitlint. Please change your commit message according to:
https://github.com/Patternslib/Patterns/blob/master/docs/developer/styleguide.md#commits-messages

2) See my code suggestion. I think it's simpler and I don't know if Backbone goes unnecessarily crazy when changing some state attributes too often. My simplification assigns the value in one go.

petschki:
Looking at the issue, I'd recomment to completely remove the pagination when there is no result in the listing. Would make more sense to me. Btw. thank you @devdanzin for contributing! We were talking briefly at the conference about the `plone.protect` issue. Very much appreciated!

devdanzin:
Thank you very much for the review!

>     1. We're using automatic changelog generation based on commit messages

I've adapted your suggestion, please let me know if it should be improved.

>     2. See my code suggestion.

I tried @petschki's suggestion of not showing the pagination text and it looks better to me. Please let me know which option we should go with and whether the code makes sense.

devdanzin:
> I'd recomment to completely remove the pagination when there is no result in the listing.

Done, not sure whether it's better to leave the empty div or also make it conditional. I've left it in current form.

> Btw. thank you @devdanzin for contributing! We were talking briefly at the conference about the `plone.protect` issue.

Thank you for the review and for helping me at the sprint!

thet (approved):
Yes, not showing the pagination at all is even a better solution.
I lean towards not showing the div when there are no results because I favor simple, small and clean HTML output. But I'm OK with the current solution.

petschki (approved):

PR #1423, opened 2025-02-05
Prepare 5.2 final

petschki:
~~Updated `Patternslib=9.10.1-beta.2`~~

Talked with @thet and we agreed to not use patternslib beta in 5.2 final.

### DIFF
--- a/src/pat/structure/structure.test.js
+++ b/src/pat/structure/structure.test.js
@@ -623,11 +623,12 @@ describe("Structure", function () {
                 msg: "pasted",
             }));
 
-            // item pending to be pasted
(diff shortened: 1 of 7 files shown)


src/pat/structure/structure.test.js
```diff
@@ -710,7 +712,8 @@ describe("Structure", function () {
         expect(structureUrlChangedPath).toEqual("");
     });
 
-    it("test navigate to folder pop states", async function () {
+    // skip for now -> decreases test time from 130s to 30s !
```
thet, line 715: :+1:

PR #1424, opened 2025-02-11
`pat-contentbrowser` UI features: global search, level filtering, focused colum

### LINKED ISSUE #1403: allow global search in pat-contentbrowser per default
i think we should change the default search behavior of pat-contentbrowser. it should be a global search, not only a search in the last column. it's not intuitive. my editors wants a "Search" independend of the place where they stay. but it should not be a switch like in the old `pat-relateditems` pattern. that is very confusing for editors.

petschki:
fixes #1403


1letter (approved):
Thanks, LGTM. I will make an PR in Products.CMFPlone with the fixed robottests

PR #1432, opened 2025-02-14
fix language identifier for language file

1letter:
fix the generated filename for language files in select2

### DIFF
--- a/src/pat/select2/select2.js
+++ b/src/pat/select2/select2.js
@@ -115,7 +115,7 @@ export default Base.extend({
                 if (lang.split("-").length > 1) {
                     lang =
                         lang.split("-")[0] +
-                        "_" +
+                        "-" +
                         lang.split("-")[1].toUpperCase();
                 }         
                 await import(`select2/select2_locale_${lang}`);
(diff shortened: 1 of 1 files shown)


src/pat/select2/select2.js
```diff
@@ -109,8 +109,16 @@ export default Base.extend({
...
-                await import(`select2/select2_locale_${this.options.language}`);
+            if (this.options.language && this.options.language !== "en" && !this.options.language.startsWith("en")) {
+                var lang = this.options.language;
+                // Fix for country specific languages
+                if (lang.split("-").length > 1) {
+                    lang =
+                        lang.split("-")[0] +
+                        "-" +
+                        lang.split("-")[1].toUpperCase();
+                }
```
petschki, line 120: instead of calling `.split()` 3 times I'd suggest:

```suggestion
                let lang = this.options.language.split("-");
                // Fix for country specific languages
                lang = (lang.length > 1) ? `${lang[0]}-${lang[1]}` : lang[0];
```

petschki (changes requested):

petschki (approved):

PR #1437, opened 2025-02-19
Fix browsing large folders with `selectableTypes` constraint

### LINKED ISSUE #1436: `pat-contentbrowser` browsing large folders with `selectableTypes` constraint is broken
## Scenario

We have a large folder with pages and images inside. The images are at the end of the page listing.

When trying to select an image in TinyMCE, the folder is loaded, but since the images are on the end, nothing is shown. (only a flickering loader)

## Reason

The `@@getVocabulary` view loads all folder items and filters afterwards for `selectableTypes`

## Possible solution

Try to load only `selectableTypes` inside folder... but we need to show "browseable" types too in order to preserve browsing subfolders.

petschki:
fixes #1436 

### Solution

We now only query for `selectableTypes` inside a folder ... additionally there's a new `browseableTypes` (default: `["Folder"]`) pattern option which can be configured to enable browsing subitems while constraining selectable types.

This also makes browsing much faster with activated selectable types...


petschki:
Just a note for the `inPath` check:

if you hat folders like:

```
/media/images
/media/images-old
```

both were marked as selected, when you clicked "images-old" ... this is now fixed.

PR #1449, opened 2025-03-11
Improve `pat-contentbrowser` upload experience

### LINKED ISSUE #1443: `pat-contentbrowser` implement "upload and add" similar to `pat-relateditems` upload functionality
We have a folder with many items and there are some UI issues when enabling upload feature:

1. when you upload to this folder, the items are appended at the bottom, so you do not see that they are there.
2. you have to explicitely select and add the uploaded items which differs from the `pat-relateditems` workflow, where the item got added directly after the upload. This could/should be the default behavior for contentbrowser upload IMHO.

To overcome 1. we implemented a `IObjectAddedEvent` listener in our project which puts the newly added item to the top of the folder.

Note to 2.: maybe we add a button "upload only" but the upload itself defaults to adding the items to the selection.

/cc @plone/classicui-team

petschki:
fixes #1443 

in action:

https://github.com/user-attachments/assets/5ef7c070-3ad1-4b36-ad13-85a6670f6ed0

Note:

the `basePath` commit is related to this issue https://github.com/plone/Products.CMFPlone/pull/4134

### DIFF
--- a/src/pat/contentbrowser/src/ContentBrowser.svelte
+++ b/src/pat/contentbrowser/src/ContentBrowser.svelte
@@ -82,8 +82,12 @@
             relativePath: "@@fileUpload",
             allowPathSelection: false,
             hiddenInputContainer: ".upload-wrapper",
-            success: (fileUpload, obj) => {
+            success: (fileupload, obj) => {
+                $previewUids.push(obj.UID);
+            },
+            queuecomplete: (fileUpload, obj) => {
                 contentItems.get({ path: $currentPath, updateCache: true });
+                showUpload = false;
             },
         });
     }


MrTango (approved):
LGTM

1letter (approved):

petschki:
OK, the initial basePath problem needs more love ... right now its fixed to the configured base path. It should of course use the currentPath if I browse through the content...

thet (approved):
LGTM!

Regarding issue Number 1 in https://github.com/collective/collective.folderorder where you mention the folder order. But there is also this solution: https://github.com/collective/collective.folderorder
This PR does not address the folder order, which is good because it wouldn't belong here.

petschki:
@thet thank you for the hint to the addon (thats exactly what I've been looking for a long time :). I propose not to care about folder order in Classic-UI core, but we can mention that there are addons that solves the problem.

petschki:
> OK, the initial basePath problem needs more love ... right now its fixed to the configured base path. It should of course use the "currentBasePath" if I browse through the content...

I'll do a separate PR with some basePath tweaks ... let's merge this.

PR #1452, opened 2025-04-24
fix(pat tinymce): set language code for init options

1letter:
update init options for the tinymce editor, set the language code after underscore replacement

### DIFF
--- a/src/pat/tinymce/tinymce--implementation.js
+++ b/src/pat/tinymce/tinymce--implementation.js
@@ -134,6 +134,7 @@ export default class TinyMCE {
         if (lang !== "en" && self.options.tiny.language !== "en") {
             try {
                 await import(`tinymce-i18n/langs7/${lang}`);
+                self.options.tiny.language = lang;
             } catch {
                 log.debug("Could not load TinyMCE language: ", lang);
                 try {
--- a/src/pat/tinymce/tinymce--implementation.js
+++ b/src/pat/tinymce/tinymce--implementation.js
@@ -131,10 +131,12 @@ export default class TinyMCE {
                 lang.split("-")[1].toUpperCase();
         }
 
-        if (lang !== "en" && self.options.tiny.language !== "en") {
+        self.options.tiny.language = lang;
+
+        if (lang !== "en") {
+            // load translations from tinymce-i18n
             try {
                 await import(`tinymce-i18n/langs7/${lang}`);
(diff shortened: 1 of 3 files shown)


petschki (changes requested):
Can you take a look here https://github.com/plone/mockup/blob/c70992f22af594bcfb93d9496b33dba8863da285/src/pat/tinymce/tinymce.js#L125-L126 ... the `options.tiny.language` is set there too ... can you try to sanitize this?

1letter:
> Can you take a look here
> 
> https://github.com/plone/mockup/blob/c70992f22af594bcfb93d9496b33dba8863da285/src/pat/tinymce/tinymce.js#L125-L126
> 
> ... the `options.tiny.language` is set there too ... can you try to sanitize this?

I'm not sure if this really necessary.
```
this.instance = new implementation(this.el, this.options);
await this.instance.init();
```
the instance init method call `await self.initLanguage();` and set the option. if i refactor this, then we have double code. so my first thought.

petschki:
Ideally this should not be re-set in the init method. Maybe conditionally:

```javascript
    if (!this.options.tiny.language) {
        var i18n = new I18n();
        var lang = i18n.currentLanguage;
        this.options.tiny.language = lang;
    }
```

src/pat/tinymce/tinymce--implementation.js
```diff
@@ -134,6 +134,7 @@ export default class TinyMCE {
         if (lang !== "en" && self.options.tiny.language !== "en") {
```
claytonc, line 134: Shouldn't this condition be an OR or AND instead?

Because the value of self.options.tiny.language by default is "en".

It's not working for me in pt_BR language. 

After changing to OR it works.

petschki:
there`s no need for language in `this.options` since we're always negotiating the current language from the system... what do you mean?

/cc @erral adding you since you've introduced this.

petschki (approved):

erral:
When I added this I would say that TinyMce was not negotiating the language and was always showing the interface in English.

But this behavior may be now obsolete.

I haven't see any language related issues in Plone 6.0 and 6.1 at least.

1letter:
I tested with "pt-br","de","en". The UI changed the language.

petschki:
I additionally tested with `de-ch` which correclty falls back to `de` for `tinymce-i18n` ... I'll merge when green

PR #1458, opened 2025-05-08
Fix "menubar" setting

petschki:
As discussed here https://github.com/plone/plone.staticresources/issues/379 we're quickfixing the menubar value type until its fixed in `plone.base`


1letter (approved):
LGTM, thanks for the quick fix.

PR #1468, opened 2025-06-08
Make Autotoc tab markup compatible with Bootstrap

cihanandac:
This fixes first part of the [plone/Products.CMFPlone#4106](https://github.com/plone/Products.CMFPlone/issues/4106)

This change updates the markup and behavior of the AutoTOC to make it more compatible with Bootstrap's tab component. Key changes:

- Moves all fieldset elements inside a tab-content parent element.
- Replaces anchor (<a>) tab triggers with Bootstrap-compatible <button> elements, utilizes data-bs-toggle, data-bs-target, and other related attributes.
- Adds appropriate aria attributes for accessibility and correct tab behavior.
- Updates tab-click handling to use Bootstrap's tab JavaScript API, replacing the previous custom logic.

### DIFF
--- a/src/pat/autotoc/autotoc.js
+++ b/src/pat/autotoc/autotoc.js
@@ -1,6 +1,7 @@
 import $ from "jquery";
 import Base from "@patternslib/patternslib/src/core/base";
 import utils from "@patternslib/patternslib/src/core/utils";
+import Tab from "bootstrap/js/dist/tab";
 
 export default Base.extend({
     name: "autotoc",
@@ -10,8 +11,9 @@ export default Base.extend({
         section: "section",
         levels: "h1,h2,h3",
         IDPrefix: "autotoc-item-",
-        classTOCName: "autotoc-nav",
-        classSectionName: "autotoc-section",
+        classTOCName: "autotoc-nav nav",
+        classContentAreaName: "autotoc-content tab-content",
+        classSectionName: "autotoc-section tab-pane fade",
         classLevelPrefixName: "autotoc-level-",
         classActiveName: "active",
         scrollDuration: "slow",
@@ -26,23 +28,52 @@ export default Base.extend({
 
         var self = this;
 
-        self.$toc = $("<nav/>").addClass(self.options.classTOCName);
+        self.$nav = $("<nav/>")
+            .attr("aria-label", "Tab Navigation")
+
+        self.$toc = $("<ul/>")
+            .addClass(self.options.classTOCName)
+            .appendTo(self.$nav);
 
         if (self.options.prependTo) {
-            self.$toc.prependTo(self.options.prependTo);
+            self.$nav.prependTo(self.options.prependTo);
         } else if (self.options.appendTo) {
-            self.$toc.appendTo(self.options.appendTo);
+            self.$nav.appendTo(self.options.appendTo);
         } else {
-            self.$toc.prependTo(self.$el);
+            self.$nav.prependTo(self.$el);
         }
+
 
         if (self.options.className) {
             self.$el.addClass(self.options.className);
         }
 
-        $(self.options.section, self.$el).addClass(self.options.classSectionName);
+        $(self.options.section, self.$el)
+            .addClass(self.options.classSectionName)
+            .attr("role", "tabpanel")
+            .attr("tabindex", "0")
 
         var asTabs = self.$el.hasClass("autotabs");
+
+        if (asTabs) {
+            self.$toc.addClass("nav-tabs");
+            self.$contentArea = $("<div/>").addClass(self.options.classContentAreaName);
+            self.$contentArea.insertAfter(self.$nav);
+            $(self.options.section, self.$el).appendTo(self.$contentArea);
+            $(self.options.section, self.$el).find("legend").hide();
+        } else {
+            self.$toc.addClass([
+                "flex-column",
+                "float-end",
+                "border",
+                "mt-0",
+                "me-0",
+                "mb-3",
+                "ms-3",
+                "py-2",
+                "px-0"
+            ])
+        }
 
         var activeId = null;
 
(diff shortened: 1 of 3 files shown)


ale-rt (commented):
Thanks a lot @cihanandac 
This LGTM, but I leave the review to @thet and @petschki

petschki (changes requested):
I have tried this out and have found some issues:

### editform tabs

- the active tab bottom border isn't transparent anymore
- the tabs have too much horizontal spacing (compared to classic demo site)

*new:*

Bildschirmfoto 2025-06-12 um 12 45 35

*original:*

Bildschirmfoto 2025-06-12 um 12 45 42


### table of contents

the `table_of_contents` display is now a tabbed display instead of a vertical link list:

*new:*
Bildschirmfoto 2025-06-12 um 12 45 12

*original:*
Bildschirmfoto 2025-06-12 um 12 45 02

src/pat/autotoc/autotoc.js
```diff
@@ -54,39 +65,43 @@ export default Base.extend({
...
-            const $nav = $("<a/>");
-            $nav.appendTo(self.$toc)
+
+            const $navItem = $("<li/>");
+            $navItem
+                .addClass("nav-item")
+                .attr("role", "presentation")
+                .appendTo(self.$toc);
+
+            const $nav = $("<button/>");
```
petschki, line 86: I would not use `<button>` here ... according to the docs (https://getbootstrap.com/docs/5.3/components/navs-tabs/#tabs) we could leave them `a` tags ...

petschki:
One more thing: if I edit a doc, the focus is now on the first tab instead of the first input element of the form. Not sure who is responsible for that... maybe because of the `<button>` implementation -> check `pat-formautofocus` ...
Bildschirmfoto 2025-06-12 um 12 55 23

petschki (changes requested):
@cihanandac thanks for taking care of the Bootstrap tabs here, but I'd really like to cut this a bit more.

You already use the Bootstrap `Tab` library and the `ul -> li -> a` markup with its css classes correctly. So I think we should try to remove our custom css code (in `autotoc.scss`) completely and go with all the standard Bootstrap stylings. Also the right floated vertical autotocs could be solved with Bootstraps `float-end` or even responsive with `float-md-end` css-classes. I think we can cut this `pat-autotoc` pattern to a minimum here. What you think?

cihanandac:
@petschki I agree with you; there are many redundant or overwritten parts of the styling and It is much better/clear to use standard Bootstrap styling instead.

I have added two more commits to the PR related to this:

1. Add "float-end", "border" classes to the tab of content version, and hide "legend" titles in the JS and remove redundant styling from the SCSS file.
2. Move rest of the paddings, margins, and text decorations to the Bootstrap classes.

Some of the margin and padding values are not possible to replicate but I have tried to protect the same stylings for all the elements as much as possible. For example, the indentation of the table of content was 1, 2, and 3 rem, but bootstrap classes doesn't have 2 rem out of box so currently it is 1, 1.5, and 3 rem.

Please let me know about your thoughts if we should either keep both commits or leave the styling of margins/paddings same by keeping it on the SCSS file.

src/pat/autotoc/autotoc.js
```diff
@@ -36,14 +38,38 @@ export default Base.extend({
...
-        $(self.options.section, self.$el).addClass(self.options.classSectionName);
+        $(self.options.section, self.$el)
+            .addClass(self.options.classSectionName)
+            .attr("role", "tabpanel")
+            .attr("tabindex", "0")
 
         var asTabs = self.$el.hasClass("autotabs");
 
+        if (asTabs) {
+            self.$toc.addClass(["nav-tabs", "p-0", "m-0"]);
```
petschki, line 54: no need for `p-0` and `m-0`

src/pat/autotoc/autotoc.js
```diff
@@ -107,10 +137,28 @@ export default Base.extend({
...
                 });
+            if (asTabs) {
+                $nav.addClass([
+                    "mt-0",
+                    "me-2",
+                    "ms-2",
+                    "py-2",
+                    "px-3",
+                    "border-bottom-0",
+                ]);
```
petschki, line 152: This is not needed because the Bootstrap `.nav-tabs .nav-link` styles take care of that. Just keep the definitions below with `if (!asTabs)` ...

src/pat/autotoc/autotoc.js
```diff
@@ -107,10 +137,28 @@ export default Base.extend({
...
+            }else{
+                $nav.addClass([
+                    "text-decoration-underline",
+                    "p-0",
+                    "my-0",
+                    "me-3",
+                    `ms-${self.getLevel($level)+2}`
```
petschki, line 159: As you've mentioned correctly, our indentation is custom CSS, so I would add the `autotoc-level-x ` classes here explicitly and add keep css definitions in the `autotoc.scss` file. The base margin class would be `mx-3` (`my` is not relevant)

src/pat/autotoc/autotoc.js
```diff
@@ -121,15 +169,17 @@ export default Base.extend({
...
-                skipHash: true,
-            });
+            const firstTabButton = self.$toc.find("a").first()[0];
+            if (firstTabButton) {
+                const tab = Tab.getOrCreateInstance(firstTabButton);
+                tab.show();
+            }
         }
 
         // After DOM tree is built, initialize eventual validation
```
petschki, line 185: you do not have to pass `self.$el` as parameter. Every pattern has `this.$el` (jQuery) and `this.el` (DOM native) available. So you can change line 190 to `const el = this.el`

cihanandac:
@petschki Thanks so much for the suggestions. I have applied your last suggested changes and rebased all them onto the master branch. 

Also, thanks for the tip about "git commit -n", I will use it next time :)

src/pat/autotoc/autotoc.js
```diff
@@ -26,7 +28,9 @@ export default Base.extend({
 
         var self = this;
 
-        self.$toc = $("<nav/>").addClass(self.options.classTOCName);
+        self.$toc = $("<ul/>")
```
thet, line 31: IMO we really should wrap the whole structure in a `<nav>` element (note: In the toolbar the `<nav>` also vanished and we should bring it back).

This is an accessible recommendation I got from chatgpt after asking about why I see so much examples without `<nav>` elements:

```html
<nav role="tablist" aria-label="Tab Navigation">
  <ul class="nav nav-tabs">
    <li class="nav-item" role="presentation">
      <a class="nav-link active" id="tab1-tab" data-bs-toggle="tab" href="#tab1" role="tab" aria-controls="tab1" aria-selected="true">Tab 1</a>
    </li>
    <li class="nav-item" role="presentation">
      <a class="nav-link" id="tab2-tab" data-bs-toggle="tab" href="#tab2" role="tab" aria-controls="tab2" aria-selected="false">Tab 2</a>
    </li>
  </ul>
</nav>
```

This is pretty close to what your PR is doing except for everything within the `<nav>` which also holds the `role` and `aria-label` attributes.

thet (changes requested):
Thanks for the PR :tada: 

But please bring back the `<nav>` element.

Other than that, there is a lot of DOM manipulation going on here. I'd prefer having that in templates, e.g. the backend-rendered forms. But I think that's going beyond the scope of this PR and off the track of the original autotoc concept.

So fine for me with the `<nav>` introduced back.

src/pat/autotoc/autotoc.js
cihanandac: @thet Thanks for the review; it is indeed important to keep things with semantic elements. I have add a new commit that wraps the whole structure in a `<nav>` element as it is in your example.

Note: The only difference is the `role="tablist"` attribute. I have realized that this role is added to the element automatically by Bootstrap probably with the `nav` class name. Please let me know if there is a workaround for it or if this version is acceptable.

thet (approved):

petschki (approved):
Thanks! 🎉

PR #1475, opened 2025-06-20
fix(pat tinymce): load plugin emoticons correctly

1letter:


1letter:
Screenshot from 2025-06-20 23-22-20
 i know it was working... but in the moment it dont work. wired

Issue #1242, opened 2022-11-21
 Non-unique ids in Theming Controlpanel 


me-kell:
The "Theming Controlpanel" (aka. "Theme settings") has two levels of `nav`-elements with non-unique `id`s.

- 1. level: `autotoc-item-autotoc-0`, `autotoc-item-autotoc-1`
- 2. level: `autotoc-item-autotoc-0`, `autotoc-item-autotoc-1`, `autotoc-item-autotoc-2`

The given anchors are also non-unique (see `href` attributes). It is not possible to properly access the elements via a `url` with `anchor`.

I'd suggest to change the 2. level `id`s to something like `autotoc-item-autotoc-1-0`, `autotoc-item-autotoc-1-1`, `autotoc-item-autotoc-1-2`

In which repository should I post an issue?

For more details and an image see [this issue](https://community.plone.org/t/non-unique-ids-in-theming-controlpanel/15887?u=mekell) and the comments.

me-kell:
Actually both following problems should be also taken into account:

1. Requesting `/@@theming-controlpanel#autotoc-item-autotoc-2` doesn't position the form in "Custom Styles" as expected.
2. Wenn the "Save"-Button in "Custom Styles" is pressed, the form is requested again and the page goes to "Themes". This requires the user to click "Advanced setting" and "Custom Styles" again every time it saves it.

PR #1479, opened 2025-06-23
Fix `pat-autotoc` fieldset ID and refactor TOC navigation

petschki:
@cihanandac I had to refactor the id generation because I've overseen that the PR #1468 rewrites the `fieldset` ids with auto-generated values (which we only want if we have vertical TOC or colliding fieldset Ids) ... see https://jenkins.plone.org/job/pull-request-6.1-3.10/974/ for robot test details, which depend on various fieldset Ids (eg `#fieldset-settings`)

### DIFF
--- a/src/pat/autotoc/autotoc.js
+++ b/src/pat/autotoc/autotoc.js
@@ -98,7 +98,10 @@ export default Base.extend({
                 }
             }
 
-            if (activeId === null && (tabHash.indexOf(window.location.hash) != -1 || $level.hasClass(self.options.classActiveName))) {
+            // NOTE: if you have nested autotocs then you have to add the
+            // parent autotoc tabId to `options.IDPrefix` of the sub autotoc
+            // in order to mark parent and sub tab as active.
+            if (activeId === null && (window.location.hash.indexOf(tabHash) == 0 || $level.hasClass(self.options.classActiveName))) {
                 activeId = tabId;
             }
 
--- a/src/pat/autotoc/autotoc.test.js
+++ b/src/pat/autotoc/autotoc.test.js
@@ -121,7 +121,7 @@ describe("1 - AutoTOC", function () {
         registry.scan(this.$el);
         await utils.timeout(1);
 
-        expect($("> nav > ul > li > a.active", this.$el).text()).toEqual("Title 1");
+        expect($("> nav > ul > li > a.active", this.$el).text()).toEqual("Title 2");
     });
     it("custom className", async function () {
         this.$el.attr("data-pat-autotoc", "className:SOMETHING");


src/pat/autotoc/autotoc.js
```diff
@@ -78,16 +78,21 @@ export default Base.extend({
...
-            if (!id || $("#" + id).length > 0) {
+            if (!asTabs || !id || $(`#${id}`).length > 1) {
                 id = self.options.IDPrefix + self.name + "-" + i;
             }
 
             const tabId = `${id}-tab`;
-            $(section).attr("id", id).attr("aria-labelledby", tabId);
+
+            if (!asTabs) {
+                $level.attr("id", id).attr("aria-labelledby", tabId);
```
petschki, line 92: I set the auto generated ID to the level item, so that its uniquely labelled. In theory, we do not have to trigger the click event anymore, because the URL hash points to the level element. What you think @thet ?

src/pat/autotoc/autotoc.js
```diff
@@ -78,16 +78,21 @@ export default Base.extend({
         var activeId = null;
 
         $(self.options.levels, self.$el).each(function (i) {
-            const section = this.closest(self.options.section);
             const $level = $(this);
-            let id = $level.prop("id") ? $level.prop("id") : $(section).prop("id");
+            const $section = $level.closest(self.options.section);
+            let id = $section.prop("id") || $level.prop("id");
 
-            if (!id || $("#" + id).length > 0) {
```
petschki, line 85: @thet I do not understand the logic behind this. We should not auto-create an ID if we have a unique fieldset#id but for my understanding this if is always true (no ID or an element with ID) ... not really sure how this could work before.

Additionally we can generate this always if we create a vertical TOC and assign it to the level elements to get a unique URL hash.

cihanandac:
I was also not aware that there are tests that are checking for the fieldset ids :) 
I have tried this version and it seems like it correctly matches the tab ids with the fieldset counter parts. 

Also, refactored parts looks good and easier to read.

Thanks @petschki

petschki:
OK, somehow this autotoc implementation confused me ... sorted some things for clarification:

The old "autotab" implementation iterates over the `levels` selector, generates a `nav > a` tab element with a generated unique ID and adds a click event which is linked to the corresponding section visibility. So the "tab#id" and the "section#id" are completely separated of each other. Furthermore the `section#id` was untouched (this is mandatory for some robottests!) The generated tab markup looked like this:

Bildschirmfoto 2025-06-23 um 15 49 04

A bit strange, that the `href` of each tab points to its own id.

Now we have the Bootstrap markup implementation (https://getbootstrap.com/docs/5.3/components/navs-tabs/#javascript-behavior) ... we need to add things like `data-bs-` and `aria-` attributes so that we can use `Bootstrap.Tab` library.

So this markup

```html
<form>
<fieldset id="fieldset-default">
<legend id="fieldsetlegend-default">Default</legend>
</fieldset>
</form>
```

should end in

```html
<form>
<nav>
<ul class="nav nav-tabs" id="myTab" role="tablist">
  <li class="nav-item" role="presentation">
    <a href="#fieldset-default" class="nav-link active" id="autotoc-item-autotoc-0-tab" data-bs-toggle="tab" data-bs-target="#fieldset-default"  role="tab" aria-controls="fieldset-default" aria-selected="true">Default</button>
  </li>
</ul>
</nav>
<div class="tab-content" id="myTabContent">
  <fieldset class="tab-pane fade show active" id="fieldset-default" role="tabpanel" aria-labelledby="autotoc-item-autotoc-0-tab" tabindex="0">...</div>
</div>
<form>
```

Thats straight forward, but currently, there are also robottests which rely on the "old" autogenerated paneID selector (see https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/tests/robot/test_contentbrowser.robot#L126) which needs to be changed accordingly (/cc @1letter)

Theres one more thing: this old issue with the nested tabs and non-unique ID problem https://github.com/plone/mockup/issues/1242 ... maybe someone has an idea how we could handle this?

petschki:
I have now reverted the autotoc merge and released mockup 5.4.2 ... this branch is now rebased with the changes from #1468 and I will continue to work on the autotoc fixes.

staticresources with the tinymce fixes from #1476 will get released soon when https://github.com/plone/plone.staticresources/pull/385 is green.

src/pat/autotoc/autotoc.js
```diff
@@ -98,7 +98,10 @@ export default Base.extend({
                 }
             }
 
-            if (activeId === null && (tabHash.indexOf(window.location.hash) != -1 || $level.hasClass(self.options.classActiveName))) {
+            // NOTE: if you have nested autotocs then you have to add the
+            // parent autotoc tabId to `options.IDPrefix` of the sub autotoc
+            // in order to mark parent and sub tab as active.
+            if (activeId === null && (window.location.hash.indexOf(tabHash) == 0 || $level.hasClass(self.options.classActiveName))) {
```
petschki, line 104: @cihanandac this solves the nested tab activation problem. however, we need to fix the `IDPrefix` of theming controlpanel before we merge this, otherwise you cannot access the subtabs anymore.

PR #1485, opened 2025-07-16
 fix `pat-contentbrowser` physicalPath generation when a Plone site is in a sub folder structure

### LINKED ISSUE #1484: `pat-contentbrowser` not working, when Plone site is inside a folder
Curently we are not able to create a Plone site inside a folder (or ZODB mountpoint) OOTB, but there are setups which might need this (see https://github.com/plone/Products.CMFPlone/issues/4202) ... for testing, you can create a Plone site on the Zope root, cut/paste it into a Zope folder and recreate the catalog.

Now when opening the contentbrowser you (at least I) see the rootPath listing correctly, but when you want to browse into a folder, the columns are empty. This is likely because of some wrong asumptions during caluclating the context path.

/cc @ewohnlich

petschki:
fixes #1484

### DIFF
--- a/src/pat/contentbrowser/src/ContentStore.js
+++ b/src/pat/contentbrowser/src/ContentStore.js
@@ -39,24 +39,19 @@ export default function (config, pathCache) {
         if (!physicalPath.startsWith(rootPath)) {
             // The path from the returned items from "vocabularyUrl" are starting
             // relative from the Plone Site. So we need to generate the phyiscalPath here.
-            if (rootPathParts.length === 1) {
-                physicalPath = rootPath + physicalPath;
-            } else {
-                // We also have to merge the rootPath and the clicked path correctly for example:
-                // rootPath: /Plone/media, clicked path: /media/subfolder
-                // has to become:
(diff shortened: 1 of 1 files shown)


ewohnlich:
I might not be the best to review this, I'm on Windows. Even running the scripts without make I'm having problems with Yarn that is almost surely due to my environment and not a change here. If you have a branch of plone.staticresources where these results are committed I'd be happy to look at that.

petschki:
Testing mockup fixes is not really well documented. Your approach looks pretty simple. I always forget about the `bundle-plone` script, wich copies the compiled bundle over to `plone.staticresources` (assuming they are on the same directory level, like it is in `buildout.coredev` which is my base development environment).

Just to outline another approach because that's what I do in staticresource PRs to test mockup fixes on jenkins before they geht released:

1. checkout `plone.staticresources`
2. change `@plone/mockup` version in `packages.json` to the fixing branch: `"@plone/mockup": "plone/mockup.git#fix-1484"`
3. run `make all` inside `plone.staticresources`
4. pip install and run wsgi

PR #1499, opened 2025-09-04
check browseable items automatically with "is_folderish" metadata

### LINKED ISSUE #1497: Can't add an image that is in a custom folderish content type in a multilingual site
**To Reproduce**

1. Go to https://classic.demo.plone.org
2. Create a new Folderish content type (f.ex: name it "Library")
3. Go to the root of the site and create this "Library" item
4. Go inside the library and create a "Page"
5. Go inside the library and create an "Image"
6. Go to a language root folder, and create a Page:
  6.1. Add a link to the "Page" that is inside the Library: OK
  6.2. Try to add the image inside the Library to the page: **you can't browse inside the Library object**

**Expected behavior**

I would expect to browse inside the Library object to look for the Image, the same way I browse to add a link.

**Screenshots**

Adding a link:

Image

Adding an image:

Image

I would say that this is somehow related to https://github.com/plone/mockup/issues/1447 and https://github.com/plone/mockup/pull/1448

petschki:
This tries to fix #1497 

How its done:

1. every item is fetched from each level.
2. if `selectableTypes` is configured, only those are shown in the list
3. also items with `is_folderish=True` are shown and are browseable (and of course selectable too if configured)

### Important:

**This has one downside which we have to solve: batching!**

If you have a folder with 50 pages and one image at the end and you configure `selectableTypes: ["Image"]` you see an empty list, because the catalog query only fetches the first 20 items.

I can now remember, that this was the reason why I've implemented "browseableTypes" in order to contrain the catalog query to selectable *and* browseable types to fix this batching issue.

Now we have to have another strategy when you want to show only the selectable types in a large folder with many other types in it. Ideas welcome

/cc @erral @MrTango @thet @1letter @yurj


erral:
I have tested this PR and it works as expected.

I don't have such large image folders so I don't know how can I test it... I will try to create dummy folders here..

Anyway, perhaps this is another issue, but wouldn't ordering items alphabetically (using `sortable_title`) make the content browser more usable?

MrTango:
I don't understand why we are not filtering for browsableTypes in the query, what is the problem with that? If nothing is set, we skip that filter, even if selectableTypes are given. selectableTypes should only have an affected if we can select them, they should still be visible. 

The only problem I see with the current behavior, is that providing selectableTypes, is also activating the filter for browsableTypes. 

When i provide browsableTypes, then i have to put all my folderish types. I could live with that. 

I honestly rarely use the browsableTypes filter. But selectableTypes all the time. 

Maybe instead of browsableTypes, a browsableTypesBlacklist does more sense if at all.

src/pat/contentbrowser/src/ContentBrowser.svelte
```diff
@@ -95,11 +95,20 @@
         });
     }
 
+    function isBrowseable(item) {
+        return (
+            item.is_folderish &&
+            (($config.browseableTypes.length &&
+                $config.browseableTypes.indexOf(item.portal_type) != -1) ||
+                !$config.browseableTypes.length)
```
mauritsvanrees, line 103: I did not try it, but this is shorter and should have the same effect:

```suggestion
            (!config.browseableTypes.length ||
                $config.browseableTypes.indexOf(item.portal_type) != -1)
```

petschki: changed that ... thanks for the suggestion!

petschki:
### UPDATE

After talking to @MrTango we came to the following conclusion:

Loading the unbatched and filtered "selectableTypes" list would potentially lead to long loading time for large folders and especially in slow network conditions.

That's why we are now loading the batched level items as usual and in case of defined `selectableTypes` we "visually deactivate" the unselectable items.
This would look like the following screenshot for a folder with many images and files while only images are defined as selectable types (in tinymce for example):

Bildschirmfoto 2025-10-01 um 13 55 29

mauritsvanrees (approved):
This makes lots of sense.  Works for me when I try it in coredev 6.1.  Tested in a multilingual site by editing `plone/app/relationfield/behavior.py` and adding these pattern options to the related items field:

```
            "browseableTypes": ["Plone Site", "LRF"],
            "selectableTypes": ["Document"],
```

I wonder if someone will want the old way back, so that non-selectable/non-browsable items are not shown.  If you have a folder with 5000 pages and 5 Images and you only allow selecting images, the previous behaviour is nicer.  Theoretically I guess we could add a pattern option `showNonMatchingItems` or something like that, default True.  If people prefer the old way, then they could set it to False in their pattern options.  Or would that slow down the page again because filtering is needed?

If such an option is easy to add, it would be nice to have.  I myself likely won't need it though.

For me this is approved.  Thanks!

I won't merge though, I try not to interfere too much in the true ClassicUI parts of Plone. ;-)  Best to have another approval.

PR #1502, opened 2025-09-17
Use POST request for quering selected items

petschki:
We've got `414 too long URL` errors when there are many items in a relationlist field. 

Additional Fix: the selected items got cut to 100 items (batchsize) after saving/re-editing ... this is fixed here too.


1letter (approved):
LGTM, works here in my local dev system with 1000 related items. Thanks for work!

PR #1506, opened 2025-10-09
fix(pat-tinymce): Ingore english language variant translations.

thet:
For english translations, ignore country specific language varants. This fixes a probem which occured with a language variant as portal language.

TODO: verify! but I can't remember anymore.

### DIFF
--- a/src/pat/tinymce/tinymce--implementation.js
+++ b/src/pat/tinymce/tinymce--implementation.js
@@ -303,7 +303,7 @@ export default class TinyMCE {
             lang = lang.split("-")[0] + "_" + lang.split("-")[1].toUpperCase();
         }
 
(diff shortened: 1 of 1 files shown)


thet:
Something has changed in TinyMCE lanugage loading.
When loading the language js file a `tinymce.addI18n` method must already exist.
Maybe we can first initialize TinyMCE with english language and then load the translation file?
Or we use the `language_url` option?
Anyways, TinyMCE doesn't make it easy to not use their premium plugins.
https://www.tiny.cloud/docs/tinymce/latest/ui-localization/

petschki:
Oh yes. Free Tiny gets more and more cut in abilities. There is a hint to platejs in the bucharest sprint docs. Definitely worth a look for classic ui too.

PR #1509, opened 2025-10-15
"navigate" event support in toolbar and navigationmarker + other stuff

thet:
Already a long running and massive PR.
I use it for a ajaxified ClassicUI Plone site and I think it's useful in general.

More on that later, but this is also covered at the "Ajaxify Plone" talk at Ploneconf 2025.

TODO:
- Split into smaller PRs
- Tests

### DIFF
--- a/src/pat/navigationmarker/navigationmarker.js
+++ b/src/pat/navigationmarker/navigationmarker.js
@@ -27,7 +27,10 @@ class Pattern extends BasePattern {
         this.scan_navigation();
     }
 
-    scan_navigation() {
+    scan_navigation(e) {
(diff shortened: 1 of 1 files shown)


src/pat/base-url/base-url.js
```diff
@@ -0,0 +1,49 @@
...
+
+class Pattern extends BasePattern {
+    static name = "pat-base-url";
+    static trigger = "body";
+
+    init() {
+        events.add_event_listener(
+            window.navigation,
+            "navigate",
+            "thet-base-url--set",
```
petschki:
```suggestion
            "mockup-base-url--set",
```

generalize event name

PR #1566, opened 2026-03-24
feat(pat-base-url): Add new pattern to update data-base-url on the bo…

thet:
…dy tag when the URL changes.

Motivation see the README.md in this PR.


petschki (approved):
LGTM

PR #1582, opened 2026-05-05
Enable vertical scrolling if toolbar height is too low

### LINKED ISSUE #3908: Plone classic side toolbar has no scroll on small height screens
In Plone 5, on smaller screens the toolbar would be given a fixed height, overflow: auto so that it could be scrolled, and it had <li> buttons to assist.

In Plone 6 classic we do not have that. It looks like the toolbar will collapse to 'mobile' mode if the width is below a threshold, but not the height. Also if more actions are added to the toolbar it is not possible to reach them in mobile mode.

Example 1: wide, short screen. No way to reach elements below Display
image

Example 2: Adding another action in classic.demo.plone.org pushes user actions below the screen (beneath the sharing icon)
image

I used https://classic.demo.plone.org/en to test, Plone 6.0.9

petschki:
fixes https://github.com/plone/Products.CMFPlone/issues/3908

### DIFF
--- a/src/pat/toolbar/toolbar.js
+++ b/src/pat/toolbar/toolbar.js
@@ -10,7 +10,19 @@ export default Base.extend({
     parser: "mockup",
     defaults: {},
 
+    _fixDropdownStrategy: function () {
+        // Use Popper.js 'fixed' strategy for dropdowns so they are not clipped
+        // by the overflow-y: auto scroll container on #edit-zone.
+        this.$el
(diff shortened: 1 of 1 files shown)


petschki:
Background: since CSS only does not work (see https://github.com/plone/plonetheme.barceloneta/pull/432) because of Bootstrap dropdown popper config, I use the pattern code to set all the styles and configs at one place.

PR #1590, opened 2026-05-26
`pat-toolbar`: Fix hidden dropdowns when toolbar position top

### LINKED ISSUE #4329: Plone 6.2.0 Classic UI toolbar on top not working on Safari
I already mentioned this to @petschki, but let me open an issue for clarity.

* Plone 6.2.0
* Classic UI site
* In Site Settings change the toolbar to show on the top, instead of on the side.
* Go to the homepage
* In Safari, try to click one of the drop-down of the toolbar: does not work.
* Try it in Firefox or Chrome: no problem.
* If the toolbar is on the side, it works fine in all three browsers.

Furthermore, if the toolbar is on the side, and the window is so low that there is not enough room for the toolbar so you would need to scroll, the menus again don't work in Safari.  Other browsers are fine.

petschki:
- [x] I signed and returned the [Plone Contributor Agreement](https://plone.org/foundation/contributors-agreement), and received and accepted an invitation to join a team in the Plone GitHub organization.
- [x] I verified there aren't any other open pull requests for the same change.
- [x] I followed the guidelines in [Contributing to Plone](https://6.docs.plone.org/contributing/index.html).
- [x] I successfully ran code quality checks on my changes locally.
- [x] I successfully ran tests on my changes locally.
- [x] I included a [change log entry](https://6.docs.plone.org/contributing/index.html#contributing-change-log-label) in my commits.

fixes https://github.com/plone/Products.CMFPlone/issues/4329

### DIFF
--- a/src/pat/toolbar/toolbar.js
+++ b/src/pat/toolbar/toolbar.js
@@ -11,6 +11,13 @@ export default Base.extend({
     defaults: {},
 
     _fixDropdownStrategy: function () {
+        // bypass if toolbar position top:
+        if (document.body.classList.contains("plone-toolbar-top")) return;
+
+        // set container overflow
+        this.el.style.overflowY = "auto";
+        this.el.style.overflowX = "hidden";
+
         // Use Popper.js 'fixed' strategy for dropdowns so they are not clipped
         // by the overflow-y: auto scroll container on #edit-zone.
         this.$el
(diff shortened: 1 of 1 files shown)


mauritsvanrees (approved):
Yes, this works.  Thanks!

Note: with the toolbar on the side in a low window, the menus still don't open in Safari, but that may be a different issue.

petschki:
The toolbar on the side is still an issue but I need to find out why this `data-bs-popper-config='{"strategy": "fixed"}'` doesn't work correctly in Safari.

PR #1592, opened 2026-05-27
Feature `pat-filemanager`

MrTango:

-----

Closes #


MrTango:
This is close to being ready. Only missig feature is the reordering function from pat-structure, which will be added by @petschki. Then we can release this with the next plone.staticresources release, so people can start using it in Plone 6 project, by overriding the folder_contents.pt:

```
<metal:main fill-slot="main">
        <span tal:replace="structure context/@@authenticator/authenticator"/>
        <div class="pat-filemanager"
          tal:attributes="data-pat-filemanager view/options" />
</metal:main>
```

petschki:
Todays Train Session:

### Rearrange implementation
- like `pat-structure` had

### URL Sync on Folder Navigation
- Browser URL is updated via `history.pushState` when navigating into subfolders (the view suffix e.g. `/folder_contents` is preserved → reload works correctly)
- Browser Back/Forward navigates the folder structure correctly via `popstate`

### Link Integrity Check on Delete
- The `DELETE` action calls `GET /@linkintegrity?uids=…` against the **portal root** before deleting (not the current context — the endpoint is only registered for `IPloneSiteRoot`)
- **With breaches:** New `LinkIntegrityForm` modal lists affected items with their incoming links; "Delete anyway" proceeds with deletion
- **Without breaches:** `window.confirm()` — for folders, also shows the subitem count: _"Delete 1 item (including 42 subitems)?"_
- `items_total` from the API response is shown per item in the modal and surfaced as a combined `subItemsTotal` in both confirmation paths

### UI Polishing
- **Content-type icons:** All items in table and grid view now use icons from the Plone registry (`contenttype/{normalized_type}`) instead of emoji
- **"Up to parent" in table view:** First row with dashed bottom border, drop-target support (mirrors grid view)
- **"Up to parent" icon:** Bootstrap icon `arrow-90deg-left` (SVG, 3rem in grid cards)
- **Rearrange button:** Moved before the action group in the toolbar
- **Rearrange fields:** Only `Title` and `ID` (matches pat-structure defaults)
- **Sticky toolbar:** Action bar sticks to the top of the viewport while scrolling (`position: sticky`) so toolbar actions remain accessible in long folder listings
- **Uniform font size:** `--filemanager-ui-size: 0.875rem` applied to toolbar, table headers and pagination
- **Drop-target border:** Replaced `box-shadow` with directional borders — single unbroken rectangle with no cell divisions
- **Bootstrap CSS variable mapping:** All colours/spacing use `var(--bs-*)` with fallbacks; label checkboxes/radios use selector `label.filemanager-field-check` for specificity (overrides Bootstrap's `label:has(input)`)


### current state

| Structure Pattern | Filemanager |
|--------|-------|
| Bildschirmfoto 2026-05-31 um 15 56 11 | Bildschirmfoto 2026-05-31 um 16 34 53 |

petschki:
### feature implementation: Folder drop → recreate structure & upload contents

Dropping an OS **folder** (onto the list, a subfolder row, or the "up one level" card) now recreates the folder tree in Plone and uploads all nested files — after a **preview + approval** step. Plain file drops are unchanged.

**Why:** `dataTransfer.files` silently omits directories, so dragging a folder uploaded nothing. Now read via the `webkitGetAsEntry()` entries API.

**Changes**
- `utils/dropentries.ts` (new): walks dropped dirs into a manifest (files+paths, folders parents-first, counts/size); drains paginated `readEntries()`; falls back to flat upload without the API.
- `api/upload.js createFolder()` + `UploadStore.uploadTree()`: create folders parents-first, upload files into them (reusing the progress panel); create-failures don't abort the batch.
- `FolderDropStore` + `FolderDropPreview.svelte` (new): approval dialog (à la `ConfirmStore`) showing summary + folder tree.
- `ListInteractions.handleExternalDrop()`: single orchestrator for all external drops.
- New `folderType` option (`folder-type`, default `"Folder"`).

**Tests/build:** new tests for dropentries / uploadTree / createFolder / FolderDropStore; full suite green; dev build compiles.

**Note:** folder reading works on drag&drop only, not the upload button (`webkitdirectory` would be a separate change).

### Real life test:

drag a folder with 4k subfolders and 2GB data to the canvas:

Bildschirmfoto 2026-06-01 um 15 59 12

petschki (approved):
From my point of view this is ready for a 5.7.0-alpha.0 release ... since it doesn't break anything else, a feature release would be enough for me.

/cc @plone/blicca-team

1letter:
I will test it tomorrow.

petschki:
> I will test it tomorrow.

We've tested this by replacing `pat-structure` with `pat-filemanager` in `plone.app.content.browser.contents.templates.folder_contents.pt`

thet (approved):
It's hard to review in-depth, overall the code looks good structured and the functionality is impressive! :tada:

The main question is - would this massive app be something for a separate Pattern?
But that would have downsides too - it would be harder to reuse some of the code in other Patterns. Restructuring would immediately mean to touch two different projects - mockup and pat-filemanger.

Another question is how to proceed with pat-structure. To not break backwards compatibility we would need to keep it in Mockup or do a new major release. For Plone 6.3 I'd definitely switch to the new pat-filemanager Pattern in plone.app.content unless some blockers show up in the alpha testing phase.
But in the end we want to remove the pat-structure pattern to get rid of the outdated backbone, etc dependencies and also get rid of the structure popup views in plone.app.content.
Maybe we should do the big cleanup for Plone 6.4.

All these questions are no blocker and can be resolved later.

So IMO good to merge and continue with the testing phase!

petschki:
My thoughts on this:

1. deliver `pat-filemanager` within mockup in a new feature release 5.7
2. switch Plone 6.3 folder_contents to the new filemanager
3. cleanup outdated pattern code in a new mockup 6.0 major release for Plone 6.4/7 (structure, relateditems, structure-updater, textareamimetypeselector and likely more ...)

thet:
We don't have a replacement for the textareamimetypeselector yet, do we?
That's also a task to tackle.

1letter:
@petschki @MrTango 

great work! amazing!

should i open for any issue in the filemanager a new issue?

image

wrong css for the cancel button in a dialog (white on light gray)

1letter:
@MrTango @petschki 

should i write some robottests?

1letter:
the drag&drop of a row is not intuitive or a little bit flaky. sometimes i can reorder and sometimes it pop up a dialog to insert the object into another

i'm not sure what are the default action for drag&drop of a row (sort, move up down, cut/insert)

see my screencast:
[Screencast from 2026-06-03 20-19-29.webm](https://github.com/user-attachments/assets/f66876ae-eeaa-4b4a-988a-0c79829d8fc5)

1letter:
image

wrong/inconsistent design of filter "delete" button

1letter:
general: UI translations for local language happen via?

1letter:
Screenshot from 2026-06-03 21-08-51

i reached a state with a wrong home link in the breadcrumb. home linked to events, not to portal root

reproducable:

1. click in the nav on "events" 
2. click in the sidebar on "content" and open the folder_contents view
3. the breadcrumb home in folder_contents point to wrong "Home"

MrTango:
> 
> wrong/inconsistent design of filter "delete" button

not sure what you mean here and also not if you are using the most recent version, since you have a plain "löschen" button there, which is not the case for me.

MrTango:
> Screenshot from 2026-06-03 21-08-51
> 
> i reached a state with a wrong home link in the breadcrumb. home linked to events, not to portal root
> 
> reproducable:
> 
>     1. click in the nav on "events"
> 
>     2. click in the sidebar on "content" and open the folder_contents view
> 
>     3. the breadcrumb home in folder_contents point to wrong "Home"

i can't reproduce this issue, for me the home points to Plone site root, in default setup and also with multi lingual setup.

1letter:
@MrTango 
Sorry, my mistake. I override the folder_contents.pt **without** `data-pat-filemanager` options attribute.

The only small error that currently happen is the styling of hover-state of a button in a modal/popup/menu . 

see: 
[Screencast from 2026-06-04 12-04-48.webm](https://github.com/user-attachments/assets/66bf6951-1fff-4b74-affe-36108b8b6fe7)

and the unstyled "delete" button next the filter settings

image

But the rest of the errors are gone with the right `data-pat-filemanager` options attribute.

When I select a large folder and upload it using drag and drop, it takes a “long” time (8–10 seconds) for the dialogue box showing the folder structure to appear. I think that takes too long; some form of feedback for the user (a loading indicator or spinner) would be necessary.

but really impressive and great work guys!

1letter:
a little missing feature:

- show a badge/icon/info for content that effective date is set and in the future (access inactive portal content)

the old "pat-structure" show a badge:

image

MrTango:
> general: UI translations for local language happen via?

the strings are in plone.app.locales, i just updated the widgets strings there and added translations for all languages with an agent, by reusing existing translations and autogenerating the rest. I'll test the german results. Other languages probably need fine tuning over time. But that happens enterily in plone.app.locales then.
see the draft here: https://github.com/collective/plone.app.locales/pull/612
needs testing and other reviews though.

MrTango:
> a little missing feature:
> 
>     * show a badge/icon/info for content that effective date is set and in the future (access inactive portal content)
> 
> 
> the old "pat-structure" show a badge:

I like the idea, can you please add this as a separate issue.

1letter:
> > a little missing feature:
> > ```
> > * show a badge/icon/info for content that effective date is set and in the future (access inactive portal content)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > the old "pat-structure" show a badge:
> 
> I like the idea, can you please add this as a separate issue.

https://github.com/plone/mockup/issues/1595

MrTango:
Regarding the size, it is already loaded async, so we will load it only when needed, not with the main bundle.
it is nopt very big for what it does: about 262 KiB, the contentbrowser has around 75KiB and TinyMCE has mutiple MB as far as i can see from the stats.

MrTango:
i fixed the drag and drop into folder, which was broken because of sortablejs preview, now preview is still used for dragging between files, but for folders it looks  a bit different. Try iy out. We can spend more time there and rebuild preview for folders similar what sortablejs does, but we can't activate the sortablejs preview for folders, or we will lose the drag into folder.

petschki:
Yes ... I like this Buschenschank momentum ... keep up the pace! ❤️ 

We need to merge this soon and work on smaller fixes afterwards, otherwise this gets too big here.

One missing thing is the "defaultpage" indicator, because there is no endpoint in restapi for that. but we can add this later too.

petschki:
> @MrTango @petschki
> 
> should i write some robottests?

I'd rather use the jest tests in this pattern to test this. We should start getting independent of those CMFPlone robottests.

webpack.config.js
petschki, line 1: ah ... thanks for that!

MrTango:
> > @MrTango @petschki
> > should i write some robottests?
> 
> I'd rather use the jest tests in this pattern to test this. We should start getting independent of those CMFPlone robottests.

indeed and instead of robottest which use playwright, i would rather use nativ playwright tests.
Since we are talking Python backend here, i would vote for the Python version of Playwright test, rather than Javascript/Typescript. But it depends a bit where those test should live. Mockup is not a real package, i don't know if we want the E2E test in there. plone.staticresources or plone.classicui (blicca) could be another place and there definitively Python based test would fit. In mockup repo both could work, since mockup is completly JS/TS anyway.

If i see this right, we have the unit test already in mockup, so we could also add the playwright test here, in the TS version. Can be created/recorded via VS Code.

MrTango:
> One missing thing is the "defaultpage" indicator, because there is no endpoint in restapi for that. but we can add this later too.

we could add the endpoint to plone.classicui and later to plone.blicca.
I have the code somewhere laying around.

davisagli:
@MrTango I would be ok with adding the default page in plone.restapi if someone else implements it. It is a feature of the backend content type system which is not specific to a particular frontend (even though Volto chooses to avoid using it).

There is an existing old PR https://github.com/plone/plone.restapi/pull/944 which maybe can be updated.

MrTango:
I'll try to look at that one, even though it covers more then just default page. but layouts sound good to have too

davisagli:
@MrTango Oh, I didn't see that that PR also has something for using a different serializer based on the selected layout. I would rather not include that (or at least consider it separately from adding the default page). The serialization already includes the name of the selected layout.

MrTango:
the way I was implementing ut, was via service endpoint if I remember correctly. that would be a small feature

davisagli:
@MrTango Maybe https://github.com/plone/plone.restapi/pull/983 is a better starting point

davisagli:
@MrTango Hmm, but for the filemanager, what you actually want is to add the is_default_page metadata column to the summary serializer in https://github.com/plone/plone.restapi/blob/main/src/plone/restapi/serializer/summary.py (we want it when serializing brains from a catalog result, not full content items)

MrTango:
> the drag&drop of a row is not intuitive or a little bit flaky. sometimes i can reorder and sometimes it pop up a dialog to insert the object into another
> 
> i'm not sure what are the default action for drag&drop of a row (sort, move up down, cut/insert)
> 

the way it should work is as follow :

drag and drop between files or folders should be possible. drag and drop into a folder as well. currently it depends where you drop it. if you drop it in the middle if the folder you will move it into the folder after confirmation. if you drop it right before or after, it will moved there. please try the current version.

MrTango:
> @MrTango Hmm, but for the filemanager, what you actually want is to add the is_default_page metadata column to the summary serializer in https://github.com/plone/plone.restapi/blob/main/src/plone/restapi/serializer/summary.py (we want it when serializing brains from a catalog result, not full content items)

true here, i was rendering complete Blicca content pull drom the restapi and needed the default page info for that. but here the meta column makes more sense. but we should probably support both. two different PRs for that?

davisagli:
@MrTango No preference, I can review those together or separate

1letter:
> > the drag&drop of a row is not intuitive or a little bit flaky. sometimes i can reorder and sometimes it pop up a dialog to insert the object into another
> > i'm not sure what are the default action for drag&drop of a row (sort, move up down, cut/insert)
> 
> the way it should work is as follow :
> 
> drag and drop between files or folders should be possible. drag and drop into a folder as well. currently it depends where you drop it. if you drop it in the middle if the folder you will move it into the folder after confirmation. if you drop it right before or after, it will moved there. please try the current version.

i tried it with the current codebase.... perfect! it works like charm!

MrTango:
@thet @petschki do you have an opinion about, where the E2E test should live: https://github.com/plone/mockup/pull/1592#issuecomment-4630090472 ?

MrTango:
I added the badges, including for working copy. I also finally hide the header parts.

grafik

petschki:
> @thet @petschki do you have an opinion about, where the E2E test should live: [#1592 (comment)](https://github.com/plone/mockup/pull/1592#issuecomment-4630090472) ?

Hm. Integration wise I think it should live in mockup, but acceptance tests for the product `Plone Blicca` could go to `plone.blicca` distribution ...

petschki:
> I added the badges, including for working copy. I also finally hide the header parts.

Nice badges... but the colors need a bit love ... let me check some color palette generators to sanitize that (like https://htmlcolorcodes.com/color-wheel/)

petschki:
> I added the badges, including for working copy. I also finally hide the header parts.

What about removing the footer too?

1letter:
The Link of Images and Files in the rows should point to the "/view"

petschki:
> The Link of Images and Files in the rows should point to the "/view"

fixed !

petschki:
I've fixes some minor things, which came up here:

- state badge color with `--plone` root var
- active button color with `--bs-primary` root var
- layout shift when "drag into folder" fixed
- workflow label translations via `plone` domain
- "house" icon for top level breadcrumb
- generate rearrange index list with querystring endpoint (sortable=True)
- button to "reset order" when column sort is clicked

petschki:
I'd say lets merge the current state and release it as feature alpha (also staticresources). I will then make a community announcement for filemanager and how to test the alpha state (by jbot-ing folder_contents.pt) to get some feedback on this.

default page endpoint and indicator can be added later ...

1letter:
> I also finally hide the header parts.

@petschki Why?

MrTango:
> I'd say lets merge the current state and release it as feature alpha (also staticresources). I will then make a community announcement for filemanager and how to test the alpha state (by jbot-ing folder_contents.pt) to get some feedback on this.
> 
> default page endpoint and indicator can be added later ...

sound good, I'll make a screen cast and go thru the different features, probably this evening.

MrTango:
I aslo would like to verify if the bug with the effective date in the futre hading the content event from owners, when they are not administrators.

petschki:
> > I also finally hide the header parts.
> 
> @petschki Why?

Well ... @MrTango and I saw that in the new Volto/Aurora folder contents implementation and we liked that... maybe we should place a title like `Filemanager` on top though ... but since we're in alpha state, things can change 😉

petschki:
> I aslo would like to verify if the bug with the effective date in the futre hading the content event from owners, when they are not administrators.

I think this is a catalog query problem and out of scope here ...

MrTango:
> > I aslo would like to verify if the bug with the effective date in the futre hading the content event from owners, when they are not administrators.
> 
> I think this is a catalog query problem and out of scope here ...

yes we don't have to wait, I'll fix this for a client soon.

MrTango:
Footer is hidden now. But both are only hidden via CSS, which is not perfect, as you can see a short flicker.

PR #1600, opened 2026-06-08
fix(pat-select2): Show predefined value for single-select widgets.

thet:
Single-select widgets did not show their predefined value. This is now fixed.

fixed:

image

not fixed:

image

-----

Closes #

### DIFF
--- a/src/pat/select2/select2.test.js
+++ b/src/pat/select2/select2.test.js
@@ -123,6 +123,27 @@ describe("Select2", function () {
         expect($(el).parent().find(".select2-chosen").text()).toEqual("Europe/Vienna");
     });
 
+    it("renders the preset value of a single-select select widget", async function () {
+        document.body.innerHTML = `
+            <select class="pat-select2">
+                <option value="Europe/Paris">Europe/Paris</option>
+                <option value="Europe/Vienna" selected>Europe/Vienna</option>
+            </select>
+        `;
+
+        registry.scan(document.body);
+        await utils.timeout(1);
+
+        const el = document.querySelector("select.pat-select2");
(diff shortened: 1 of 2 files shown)


src/pat/select2/select2.test.js
```diff
@@ -100,6 +100,31 @@ describe("Select2", function () {
         expect(select2.options.ajax.url).toEqual("select2-users-vocabulary");
     });
 
+    it("renders the preset value of a single-select vocabulary widget", async function () {
+        document.body.innerHTML = `
+            <input class="pat-select2"
```
petschki, line 105: Shouldn't that be a `<select>` if its not multiple values ? 

I wonder why this is needed, because the `language` field actually works with predefined values: https://github.com/plone/plone.app.dexterity/blob/master/src/plone/app/dexterity/behaviors/metadata.py#L110

petschki (changes requested):

src/pat/select2/select2.test.js
thet: A‌ `<select>` field cannot use AJAX‌ remote data.

And for the timezone selection fields I do not want to use pre-populated list of options - because it can be huge - but a  text field where you can enter manually (if JS is off) or search/select via AJAX.

thet:
@petschki

Btw. I have added a test showing that the `<select>` field isn't broken with this PR.
But thanks for the fixes in the querystringwidget!

src/pat/select2/select2.test.js
petschki: that makes totally sense

petschki (approved):
LGTM

PR #1605, opened 2026-06-15
fix(modal): title extraction method (jquery removed .context)

Gomez:
Fixes `JSError: Uncaught TypeError: can't access property "innerText", e.$el.context is undefined ` if `data-pat-plone-modal="image: true"`


Gomez:
Replaced: https://github.com/plone/mockup/pull/1606

PR #1606, opened 2026-06-15
fix(modal): title extraction method (jquery removed .context)

Gomez:
Fixes `JSError: Uncaught TypeError: can't access property "innerText", e.$el.context is undefined ` if `data-pat-plone-modal="image: true"`


petschki (approved):
LGTM. Thanks!

petschki:
@Gomez would be great if you could cherry pick this commit and create a second PR based on master.

PR #1608, opened 2026-06-17
feat(core): Global lazy Intl polyfills and locale-sensitive date formatting

erral:
### Summary
This PR improves date formatting across Plone patterns and introduces a global mechanism for lazily loading  polyfills and locale data.

### Key Changes
- **Locale-Sensitive Formatting**: Updated `formatDate` in `pat-filemanager` to respect the Plone site language from `<html lang='...'>` (normalized for `Intl`).
- **Lazy Intl Polyfills**: Added `src/core/intl-loader.js` which detects if the browser natively supports the current locale. If not, it lazily fetches the `@formatjs/intl-datetimeformat` polyfill and specific locale data chunks. This ensures languages like Basque (`eu`) are correctly supported even in limited environments.
- **Pattern Integration**: The polyfill loader is now part of the initialization for `pat-filemanager`, `pat-contentbrowser`, and `pat-structure`.
- **Build Support**: Configured Webpack aliases to handle dynamic imports of locale data smoothly, enabling clean builds and efficient chunking.
- **Functional Testing**: Added `src/core/intl-loader.test.js` verifying that missing locales trigger polyfill loading and result in correct language-specific output.

### Why?
Previously, date formatting often defaulted to the browser's language, leading to UI inconsistencies in Plone sites with a different language than the user's OS. Furthermore, many browsers lack data for certain languages (like `eu`), causing formatting to fallback to English or fail. This PR provides a robust, per-language lazy-loading solution that fixes these issues for all patterns with minimal overhead.

### DIFF
--- a/src/pat/filemanager/src/App.svelte
+++ b/src/pat/filemanager/src/App.svelte
@@ -122,6 +122,24 @@
 
     onMount(async () => {
         await ensureIntlSupport(getLang());
+    });
+    
+    onMount(() => {
+        contents.load();
+
+        function onPopState(event) {
+            const ctx = event.state?.contextUrl;
+            if (ctx && ctx !== contents.contextUrl) {
+                isRestoringHistory = true;
+                contents.navigateTo(ctx).finally(() => {
+                    isRestoringHistory = false;
+                });
+            }
+        }
+        window.addEventListener("popstate", onPopState);
+        return () => window.removeEventListener("popstate", onPopState);
+    });
+        await ensureIntlSupport(getLang());
         contents.load();
 
         function onPopState(event) {
--- a/src/pat/filemanager/src/App.svelte
+++ b/src/pat/filemanager/src/App.svelte
@@ -140,6 +140,24 @@
         return () => window.removeEventListener("popstate", onPopState);
     });
         await ensureIntlSupport(getLang());
+    });
+    
+    onMount(() => {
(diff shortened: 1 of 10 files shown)


src/pat/filemanager/src/App.svelte
```diff
@@ -118,7 +120,8 @@
 
     let isRestoringHistory = false;
 
-    onMount(() => {
+    onMount(async () => {
```
MrTango, line 127: this is problematic, because the existing event handler will not be cleaned up with async.

```suggestion
    onMount(async () => {
        await ensureIntlSupport(getLang());
    });
    
    onMount(() => {
        contents.load();

        function onPopState(event) {
            const ctx = event.state?.contextUrl;
            if (ctx && ctx !== contents.contextUrl) {
                isRestoringHistory = true;
                contents.navigateTo(ctx).finally(() => {
                    isRestoringHistory = false;
                });
            }
        }
        window.addEventListener("popstate", onPopState);
        return () => window.removeEventListener("popstate", onPopState);
    });
```

MrTango (changes requested):

MrTango:
@erral when you have time, please have another look at this ;)

erral:
@MrTango I have tested it and it works with your changes too.

MrTango (approved):

PR #1611, opened 2026-07-01
Fix `pat-structure` folder contents sort order [5.6.x]

petschki:
### Problem
In pat-structure (folder_contents) the listing is sorted incorrectly once a folder contains more than a few items. The server response (@@getVocabulary) is correctly ordered — only the rendered table is scrambled.

### Root cause
The table row rendering in src/pat/structure/js/views/table.js iterates the collection with an async callback that appends each row only after await view.render(). Since collection.each does not await and row rendering is async (icon resolution, action menu), rows get appended in the order their render promises resolve rather than in collection order. With few rows this is invisible; with more rows the race scrambles the DOM order, and DataTables (order: []) does not re-sort it.

### Why it only appeared in 5.6.0+
The race has existed since 5.4.x but was masked: DataTables was initialized with order: [0, "asc"], which re-sorted rows by the hidden _sort (server order) column on every draw. [#1552](https://github.com/plone/buildout.coredev/issues/1552) changed this to order: [] — correct in intent, but it removed the safety net that had been hiding the race, so the bug became visible in 5.6.0. Not reproducible in Plone 6.1 / mockup 5.4.x.

### Fix
Build the row views in collection order, render them in parallel via Promise.all, then append them in order — so the DOM order always matches the collection order, independent of any DataTables ordering setting. The now-obsolete table_row_rendering_finished event mechanism is removed.

### DIFF
--- a/src/pat/structure/js/views/table.js
+++ b/src/pat/structure/js/views/table.js
@@ -1,7 +1,6 @@
 import $ from "jquery";
 import _ from "underscore";
 import _t from "../../../../core/i18n-wrapper";
-import events from "@patternslib/patternslib/src/core/events";
 import registry from "@patternslib/patternslib/src/core/registry";
 import BaseView from "../../../../core/ui/views/base";
 import TableRowView from "./tablerow";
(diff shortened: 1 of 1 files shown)


pbauer (approved):
lgtm

thet (approved):

thet:
LGTM. But shouldn't this also applied to `master` as long as we keep pat-structure there?

petschki:
Yes of course. I'll merge this and cherry pick the commit to master later.

thet:
@petschki erster: https://github.com/plone/mockup/pull/1612

PR #1612, opened 2026-07-08
fix(pat-structure): keep table rows in collection order

thet:
Cherry-pick of https://github.com/plone/mockup/pull/1611

The table row rendering iterated the collection with an async callback (`collection.each(async ...)`) that appended each row only after awaiting `view.render()`. Row rendering is async (it awaits icon resolution), so rows were appended in the arbitrary order their render promises resolved rather than in collection order, corrupting the sort order of the folder contents.

This stayed hidden until 5.6.0: DataTables was initialized with `order: [0, "asc"]`, which re-sorted the rows by the hidden `_sort` column (the server order index) on every draw and thus masked the race. Commit 57cd7dd ("Fix ordering logic.") changed this to `order: []`, removing the safety net and exposing the underlying bug.

Fix the root cause: build the row views in collection order, render them all in parallel via `Promise.all`, then append them in order. This guarantees the DOM order matches the collection order regardless of the DataTables ordering settings. The now-obsolete
`table_row_rendering_finished` event dance and its `events` import are removed.

-----

Closes #

### DIFF
--- a/src/pat/structure/js/views/table.js
+++ b/src/pat/structure/js/views/table.js
@@ -1,7 +1,6 @@
 import $ from "jquery";
 import _ from "underscore";
 import _t from "../../../../core/i18n-wrapper";
-import events from "@patternslib/patternslib/src/core/events";
 import registry from "@patternslib/patternslib/src/core/registry";
 import BaseView from "../../../../core/ui/views/base";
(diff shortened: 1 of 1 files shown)


petschki (approved):
LHTM

PR #1613, opened 2026-07-13
Lazify pat-querystring: split registration from implementation

MrTango:
pat-querystring contributed ~33KB src to the eager patterns chunk on every page, though the widget only appears on Collection edit forms and the querystring criteria UI. It also statically imported the contentbrowser pattern, dragging an eager edge into pat-contentbrowser.

Split it into a thin registration module and querystring--implementation.js (holding the widget and the contentbrowser import). 

addresses #1618

PR #1614, opened 2026-07-13
Lazify pat-recurrence: split registration from implementation

MrTango:
pat-recurrence was the single largest pattern in the eager patterns chunk (~52KB src, incl. three statically-imported XML templates and the Modal dependency), loaded on every page though the widget only appears on Event add/edit forms.

Split it into a thin registration module (trigger, parser, and an init() that dynamically imports the implementation on first match) plus recurrence--implementation.js holding the widget, the XML templates and the Modal import. Follows the existing tinymce--implementation lazy-pattern convention. 

addresses #1618


petschki (approved):
LGTM!

changed commit message same as in #1613

petschki:
/run-coredev-6.3-robottests

PR #1615, opened 2026-07-13
Lazify pat-select2: split registration from implementation

MrTango:
pat-select2 contributed ~13KB src to the eager patterns chunk on every page, plus a lazy select2_locale_* context map baked into the eager chunk, though select2 widgets only appear on forms. (The select2 library itself was already an async chunk; only the wrapper and the locale map were eager.)

Split it into a thin registration module and select2--implementation.js holding the wrapper, the select2 library import and the select2_locale_* context. The implementation is grafted onto the single pattern instance in init(). Public options and emitted markup are unchanged.

pat-relateditems borrows select2's methods (Select2.prototype.* applied to its own instance). To keep that working without pulling select2 back into the eager chunk, relateditems now imports select2--implementation lazily in its init() and calls the borrowed methods through it. This is the coordinated change flagged between the select2 and relateditems lazification tickets.

The select2 wrapper and locale map now load only on pages with a .pat-select2 or .pat-relateditems element.

addresses #1618


petschki (approved):
LGTM!

Also the relateditems import fix 👍🏼

PR #1616, opened 2026-07-13
Lazify pat-upload: split registration from implementation

MrTango:
pat-upload contributed ~17KB src to the eager patterns chunk on every page, though the upload UI only appears in folder contents and upload dialogs.

Split it into a thin registration module and upload--implementation.js holding the jquery/underscore/i18n wiring and the lazily-loaded Dropzone integration. The implementation is grafted onto the single pattern instance in init(), so consumers that instantiate Upload directly (pat-structure, pat-relateditems, the contentbrowser app) keep working. Public options and emitted markup are unchanged.

The implementation now loads only where a .pat-upload element exists.

addresses #1618


petschki (approved):
LGTM!

petschki:
/run-coredev-6.3-robottests

PR #1617, opened 2026-07-13
Mockup lazy modal

MrTango:
**Merge after the pat-recurrence PR #1614!**

Lazify pat-modal: thin registration module + lazy jquery-form

Split the ~1000-line pat-modal implementation out of the every-page
eager patterns chunk. modal.js becomes a thin registration module that
loads modal--implementation.js on first .pat-plone-modal match; the
implementation grafts onto the instance from the prototype, since
imperative callers still do `new Modal()`.

jquery-form ($.fn.ajaxSubmit) moves from a top-level import to a lazy
`await import("jquery-form")` at its single call site in
handleFormAction, so it no longer loads on every page.

Imperative importers are repointed at modal--implementation:
controlpanels (registry / dexterity-types-listing / contentrule-elements)
and recurrence load it as part of their own lazy implementation chunks;
manageportlets is registered eagerly, so it loads the impl on demand
inside showEditPortlet. Side-effect registration imports (patterns.js,
tinymce links, structure actionmenu) keep pointing at the thin module.

Depends on the pat-recurrence lazification: recurrence is eagerly
registered, so its Modal import must live in recurrence--implementation
(a lazy chunk) for the modal body to actually leave the eager bundle.

addresses #1618


petschki (approved):
LGTM!

I've rebased this PR on top of the new #1614 commit, so once this is merged, the commit here disappears after rebasing.

PR #1622, opened 2026-07-20
fix(pat-tinymce): construct the modal implementation for the link modal

petschki:
The pat-plone-modal lazification left links.js creating its modal via registry.patterns["plone-modal"].init(...). That static initializer constructs the *registered* pattern — now the thin registration module, whose methods (show/hide/...) only appear after an async graft — while LinkModal calls this.modal.show() synchronously right after construction. First click on Insert Link/Image threw and the modal stayed broken, which also failed every robot test funneling through it (tinymce, contentbrowser, linkintegrity).

Construct modal--implementation directly instead; mockupParser reproduces the option parsing of the registry path. links.js is only reached via the lazily-loaded tinymce implementation, so the modal body stays out of the eager bundle. Add a regression test asserting the implementation exposes its methods synchronously.

### DIFF
--- a/src/pat/modal/modal--implementation.test.js
+++ b/src/pat/modal/modal--implementation.test.js
@@ -1,0 +1,23 @@
+import $ from "jquery";
+import Modal from "./modal--implementation";
+
+// The tinymce link modal (pat-tinymce js/links.js) constructs the modal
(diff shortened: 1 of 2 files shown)


petschki:
/run-coredev-6.3-robottests

petschki:
/run-coredev-6.3-robottests

PR #1633, opened 2026-09-05
feat(core): getCriterias preserves non-ASCII terms raw (encode-exactly-once)

rohnsha0:
- [x] I signed and returned the [Plone Contributor Agreement](https://plone.org/foundation/contributors-agreement), and received and accepted an invitation to join a team in the Plone GitHub organization.
- [x] I verified there aren't any other open pull requests for the same change.
- [x] I followed the guidelines in [Contributing to Plone](https://6.docs.plone.org/contributing/index.html).
- [x] I successfully ran code quality checks on my changes locally.
- [x] I successfully ran tests on my changes locally.
- [x] If needed, I added new tests for my changes.
- [x] If needed, I added [documentation](https://6.docs.plone.org/contributing/documentation/index.html) for my changes.
- [x] I included a [change log entry](https://6.docs.plone.org/contributing/index.html#contributing-change-log-label) in my commits.

-----

Closes https://github.com/plone/Products.CMFPlone/issues/4366

image

PR #1634, opened 2026-09-09
fix(pat structure): text filtering with umlauts [5.6.x]

petschki:
backport #1633


petschki:
/run-coredev-6.2-robottests

PR #1638, opened 2026-09-11
fix(pat contentbrowser): keep add-on components registered under the default key [master]

petschki:
see #1637


petschki:
/run-coredev-6.3-robottests

PR #1641, opened 2026-09-16
fix(pat-contentbrowser): don't fire change while rendering the initial selection

MrTango:
The selected-items effect wrote the initial value back to the input on mount and dispatched a change event for it, so pat-formunloadalert marked every untouched form with a content browser field as edited and warned when leaving the page.

Only write and dispatch when the value actually differs.


petschki:
/run-coredev-6.3-robottests

PR #1644, opened 2026-09-16
tech(core): load the formatjs Intl polyfill from a lazy module

### LINKED ISSUE #1640: Dynamic Intl locale-data import bloats the Plone bundle by ~770 lazy chunks (~340 MB on disk)
I've just released 5.7.0-alpha.1 and wondered why the unpacked bundle has now ~390MB instead of ~50MB (5.6.x) ... @erral @MrTango @thet the agentic observation showed this:

## Summary

`src/core/intl-loader.js` loads `@formatjs/intl-datetimeformat` locale data
through a dynamic import with a template string:

```js
await import(`@formatjs/intl-datetimeformat/locale-data/${baseLang}.js`);
```

Webpack cannot know the locale at build time, so it emits one lazy chunk for
**every** file in `node_modules/@formatjs/intl-datetimeformat/locale-data/`
(766 files). Each of those chunks is 50–270 KB because the locale data
includes the translated time zone names. Building with
`pnpm build:webpack:plone` now writes roughly 770 additional chunks plus
their source maps into `plone.staticresources`.

## Observed

Comparing the compiled `static/bundle-plone/` directory of
`plone.staticresources` built from mockup 5.6.10 with one built from the
current master (5.7.0-alpha):

| | mockup 5.6.10 | mockup 5.7.0-alpha |
|---|---|---|
| `chunks/*.js` files | 2811 | 3587 |
| `chunks/*.js` total | 20 MB | 133 MB |
| `chunks/*.map` files | 724 | 1500 |
| `chunks/*.map` total | 27 MB | 253 MB |
| chunks between 50 KB and 200 KB | 14 files, 1.3 MB | 727 files, 99.5 MB |
| chunks between 200 KB and 1 MB | 5 files, 1.7 MB | 61 files, 14.4 MB |
| `bundle-plone/` total | 49.5 MB | 388 MB |

Many of the new chunks have identical sizes (22 × 165849 bytes, 17 × 165467
bytes, 16 × 166979 bytes, 13 × 82637 bytes, …). Inspecting one of them shows
localized time zone names ("Atlantique", "Greenwich", "Rocheuses", …) and
`DateTimeFormat` / `timeZoneName` / `hourCycle` keys, i.e. the formatjs
locale data.

The chunks are only emitted since the commits of 2026-06-17 that made the
dynamic import resolvable for Webpack (`fbbdbf1e9`, `5a09b8cf6`,
`9c5107d2f`). The `resolve.alias` entry for
`@formatjs/intl-datetimeformat/locale-data` in `webpack.config.js` (marked
with a "NOTE: above doesn't work" comment) belongs to the same change.

## Impact

- Runtime is not affected: the polyfill path in `ensureIntlSupport()` is
  only taken when the browser does not support the site language natively
  (practically never in supported browsers), and even then only the single
  locale chunk is fetched. Page payload measurements on a Plone 6.3 site
  show no additional requests.
- The released package is affected: every `plone.staticresources` build
  now ships ~340 MB of chunks and source maps that are almost never used.
  This bloats the sdist/wheel, the git repository of
  `plone.staticresources` (the compiled bundle is committed there) and
  every Plone deployment.

## Proposed fix

Restrict the dynamic import to the languages Plone actually supports, e.g.
with a Webpack magic comment (the list should be derived from the languages
in `plone.app.locales`):

```js
await import(
    /* webpackInclude: /\/(ar|ca|cs|da|de|el|en|es|eu|fi|fr|hu|it|ja|nl|no|pl|pt|ro|ru|sv|tr|uk|zh)\.js$/ */
    /* webpackChunkName: "intl-locale-[request]" */
    `@formatjs/intl-datetimeformat/locale-data/${baseLang}.js`
);
```

Alternatives:

- Skip the polyfill entirely for the default language set and only keep
  it for explicitly configured languages.
- Load the locale data at runtime from a static URL instead of letting
  Webpack bundle it (`/* webpackIgnore: true */`), served from
  `plone.staticresources` or a CDN.

Independently of the chosen approach, it may be worth checking whether the
`.map` files of lazy chunks need to be committed to `plone.staticresources`
at all — they already make up more than half of the directory size in
mockup 5.6.10.

## Environment

- mockup master (5.7.0-alpha), `@patternslib/patternslib` 9.11.0-alpha.0,
  `@patternslib/dev` 4.1.0-alpha.0, `@formatjs/intl-datetimeformat` 7.4.9
- Build: `pnpm build:webpack:plone` (NODE_ENV=production, DEPLOYMENT=plone)
- Compared against `plone.staticresources` compiled from mockup 5.6.10
  (branch `thet/upgrade-mockup`, commit `c115758d5`)

MrTango:
Fixes #1640.

`core/intl-loader.js` dynamically imports `locale-data/${baseLang}.js` from the ~766-file `@formatjs/intl-datetimeformat/locale-data` directory. Webpack turns that into a context and emits one lazy chunk per file: ~110 MB of chunks plus ~220 MB of source maps, and the production build runs out of the default Node heap.

`baseLang` never carries a region or script suffix, so the ~450 variant files (`pt-BR.js`, `zh-Hans.js`, ...) were never requestable. This PR:

- constrains the context with `webpackInclude` to base-language files (`de.js`, `pt.js`, ...). Regional variants still resolve through the polyfill's own fallback (`pt-BR` → `pt`, `de-AT` → `de`), as before. Note: webpack matches `webpackInclude` against the absolute file path, so the pattern is not anchored at `./`.
- moves the polyfill-force import and the locale-data context into a separate lazy `intl-polyfill.js`, loaded only after feature detection fails, so the context map is not part of the pat-structure chunk.

Production build, before → after:

| | master | this PR |
|---|---|---|
| chunk files | 3587 | 3145 |
| locale-data chunks | 767 | 324 |
| dist size | 380 MB | 170 MB |
| bundle.min.js | 291 KB | 278 KB |

Re module federation (@petschki, @thet): MF only avoids loading a shared module twice. It does not change what webpack emits for a dynamic-import context, so the chunk count is only reduced by narrowing the context.


erral:
But this will not reduce the plone.staticresources package size issue described in #1640 right?

petschki:
I do not really understand the need of this "lazy loading" refactorings. I fact, I thought that's what module federation is made for -> to load resources only if they are needed ... or am I missing here something?

/cc @thet

thet:
The module federation machinery only makes sure that the same resource isn't loaded twice from different bundles. As a side effect it also enforces code splitting via the async import in the bundle entry point (`import("./bundle.js")`).

Loading only when it's needed is only done via async imports (`await import....`).

However, how this PR solves the issue isn't also clear to me. It looks like it's just async-importing another module where other async imports are done - the same async imports as before.

But I'll give this another thought after I have finished my current thought on a different topic and had some lunch. :exploding_head:

MrTango:
I updated the PR and the description. Still needs a bit of testing.

thet:
I got help reviewing that. The extra async import here improves some webpack build artefact internals - there is a chunk lookup table and that is moved down to the chunk and so not initially polluting the bundle size.

This is the conversation:

Please look at the latest commit and explain why the changes make loading of the intl polyfill cheaper. It was an async lazy import before and is the same afterwards, AFAIK.

> You’re right: **the polyfill code and individual locale data were already lazy-loaded.** Commit `3b8878b` improves the surrounding Webpack overhead, rather than making those imports lazy for the first time.
> 
> There are two changes:
> 
> 1. **Generate fewer locale chunks.** With `import(.../locale-data/${baseLang}.js)`, Webpack cannot infer that `baseLang` never contains a region suffix. Previously, it generated chunks for all matching files, including `pt-BR.js`, even though the code would only request `pt.js`. The new [`webpackInclude` filter](/home/_thet/data/dev/plone/patternslib/mockup/src/core/intl-polyfill.js:36) excludes those unreachable variants. This reduces build work, output size, and the chunk metadata in Webpack’s runtime. It does **not** save downloading hundreds of locale files: browsers weren’t downloading those before.
> 
> 2. **Lazy-load the locale lookup table itself.** A variable import requires a generated “context”: a lookup table mapping filenames to modules/chunks. Previously, that table accompanied `intl-loader.js` in its containing chunk, although the actual locale data remained lazy. The new [`import("./intl-polyfill.js")`](/home/_thet/data/dev/plone/patternslib/mockup/src/core/intl-loader.js:28) moves that table behind another lazy boundary. Browsers with native locale support now avoid fetching that table too.
> 
> So the benefit is **substantially cheaper builds and somewhat less JavaScript for browsers that don’t need the polyfill**. For browsers that *do* need it, the actual polyfill and selected locale remain essentially the same payload, with an additional sequential import step.
> 
> The commit reports `dist` shrinking from 380 MB to 170 MB and `bundle.min.js` shrinking by 13 KB; those are reported build measurements, not measurements I reproduced. The large `dist` reduction should not be interpreted as equivalent page-load savings.
> 
> I also checked the suspected regex escaping issue against the installed Webpack parser: the filter correctly accepts `de.js` and excludes `pt-BR.js`.

thet (approved):
LGTM!

Also I can reproduce the smaller bundle size - different chunk counts and sizes, but still a big improvement!

PR #1648, opened 2026-09-16
fix: default-export the bundle in the entry point.

thet:
This is necessary, so that the module federation initialization code can await bundle initialization.


petschki (approved):

petschki:
/run-coredev-6.2-robottests

PR #1649, opened 2026-09-17
Reduce bundle size with `nosource-source-map`

petschki:
Bundle sizes compared by configuration:
bundle-analyzer-top10-sidebar-vergleich

Note:

bundle-analyzer doesn't recognize `.map` files at all but there's an important finding: `css-loader` and `sass-loader` add CSS maps into the JS chunks when `devtools` is configured. that's why the tinymce chunk gets smaller when changing the value here.

I've added `sourceMap: false` to the `css/sass-loader` maps options explicitly to get rid of that too when using `nosource-source-map` option. see variant 4 in the screenshot.


thet (changes requested):
Great findings! Instead of adding this here, I want to move that into `@patternslib/dev` as new defaults.
I'm currently working on this.

petschki:
obsolete since `@patternslib/dev==4.2.0` see #1652

PR #1650, opened 2026-09-17
pat-structure bugfixes

thet:
- fix(pat-structure): Prevent datatables alerts in specific column layouts.

- fix(pat-structure): Scan rows when attached.
  Do a Patternslib registry scan on rows when they are attached to the
  document and not when they are created. This allows the Patternslib
  parser to pick up parent Pattern configurations. A common way is to
  define global Pattern options on the document body.


thet:
/run-coredev-6.2-robottests

src/pat/structure/js/views/tablerow.js
```diff
@@ -93,6 +94,13 @@ export default Backbone.View.extend({
         });
         $(".actionmenu-container", this.$el).append(await menuview.render());
 
+        // Patterns inherit configuration from ancestors, so initialize only
+        // after attachment. TableView scans newly inserted rows; rows rendered
+        // again after context-info updates are already in the document.
+        if (this.el.isConnected) {
+            registry.scan(this.$el);
```
petschki, line 101: Should we remove the `registry.scan` in the actionmenu per row if we do it here? See https://github.com/plone/mockup/blob/5.6.x/src/pat/structure/js/views/actionmenu.js#L104

thet: @petschki that exact line is removed - at this point the table row is not attached to the document, so if you configure a pattern via data attributes on the body, the settings do not reach the patterns in the table row. This PR fixes exactly that.

petschki: Ah sorry, didn't see it. Just remembered from our conversation yesterday 😉

petschki (approved):

PR #1651, opened 2026-09-17
pat-structure bugfixes

thet:
- fix(pat-structure): Prevent datatables alerts in specific column layouts.

- fix(pat-structure): Scan rows when attached.
  Do a Patternslib registry scan on rows when they are attached to the
  document and not when they are created. This allows the Patternslib
  parser to pick up parent Pattern configurations. A common way is to
  define global Pattern options on the document body.


thet:
/run-coredev-6.3-robottests

petschki (approved):

PR #1652, opened 2026-09-17
update `@patternslib/dev` to 4.2.0 and `Babel` to 8

petschki:
@patternslib/dev 4.2.0 ships the production source map setup this package needed (nosources-source-map, css/sass source maps off) and upgrades to Babel 8 targeting ES6+ browsers.

Babel 8 removes the allowDeclareFields and allExtensions options of @babel/preset-typescript (both are now the default behaviour), so drop them from the babel config, the svelte runes-in-module webpack rule and the jest transformer. @babel/core is declared as a devDependency, as the babel config and the jest transformer require it directly, and pinned to 8 via a pnpm override: babel-jest otherwise resolves its @babel/core peer to a second 7.x instance, under which the Babel 8 presets do not strip TypeScript and the .ts test suites fail to parse.


petschki:
/run-coredev-6.3-robottests

